PR #10, opened 2017-04-06
Add automatic release build in travis

houshengbo:
This patch adds the binary build in travis, which can be don automatically
when a tag is added.


houshengbo:
@dubeejw I had to merge this one as soon as possible, because it is unable to test if it not merged in master.

PR #13, opened 2017-04-11
Update Dockerfile

pritidesai:
Bundling externalized strings into wsk CLI binary

@dubeejw please review the changes. I haven't found this anywhere else being done so adding it back to docker file.

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -19,6 +19,10 @@ ADD . /src/github.com/openwhisk/openwhisk-cli
 RUN echo "Restoring Go dependencies"
 RUN cd /src/github.com/openwhisk/openwhisk-cli && /bin/godep restore -v
 
+# Collect all translated strings into single .go module
+RUN echo "Packaging i18n Go module"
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -19,6 +19,10 @@ ADD . /src/github.com/openwhisk/openwhisk-cli
 RUN echo "Restoring Go dependencies"
 RUN cd /src/github.com/openwhisk/openwhisk-cli && /bin/godep restore -v
 
+# Collect all translated strings into single .go module
+RUN echo "Packaging i18n Go module"
+RUN cd /src/github.com/openwhisk/openwhisk-cli && /bin/go-bindata -pkg wski18n -o wski18n/i18n_resources.go wski18n/resources
```
houshengbo: I do not think we need them any more, because i18n module is accessible in the source code. No need to build them .

houshengbo:
@pritidesai Priti, I do not think it is necessary to add the lines. Please check my comments.

Dockerfile
houshengbo: It used to be necessary to have them, because GO CLI within openwhisk is not in a standard go project structure. After we put in a standard go project structure, we do not need to build the module, which can be accessed by the source code instead. All the package imports have been changed to use the absolute paths.

PR #38, opened 2017-05-02
Add test cases regarding the action

houshengbo:


dubeejw:
@houshengbo, is this PR still valid?

PR #41, opened 2017-05-08
add travis button to readme

csantanapr:


csantanapr:
@csantanapr Do we want to add golint to the scanCode.py for all .go files?

sounds like a good idea

PR #65, opened 2017-06-23
Add the test cases for CLI

houshengbo:
This PR establishes the framework of scala test cases for openwhisk cli.
Since we only have scala test cases as the legacy, we tentatively port
the test cases from openwhisk to this repo.


rabbah:
i understand that we need to move these tests but i would rather not lose our git history with all the migrations that are happening.

houshengbo:
@rabbah It is not easy to do, because the cli tests are not located under one folder, but spreads into many folder, which also contains tests not for cli.

PR #74, opened 2017-06-24
Modifying the test cases for cli repo

houshengbo:
This PR establishes the framework of scala test cases for openwhisk cli.
Since we only have scala test cases as the legacy, we tentatively port
the test cases from openwhisk to this repo.


houshengbo:
@rabbah history kept.

PR #99, opened 2017-08-29
WIP Changes required to add service bindings to the Bluemix plugin

jessealva:
Essentially just added some publicly exported functions, which called the private versions.  This helped to have a smaller ripple effect across the code base as some of the functions that we needed to export would force changes across the entire API. 

This can still be done, but at least for the time being I'd prefer to do it this way and then fix this up (i.e. removing the duplicate methods that only have a difference in their case of the first character of the name and just making the required methods publicly exported) .   This way we can meet the deadline with less risk as there wouldn't be as many changes across the entire cli code base.

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -31,6 +31,14 @@ import (
 var client *whisk.Client
 const DefaultOpenWhiskApiPath string = "/api"
 
+func GetClient() *whisk.Client {
+    return client
+}
+
+func SetupClientConfig(cmd *cobra.Command, args []string) (error){
+    return setupClientConfig(cmd, args)
+}
+
 func setupClientConfig(cmd *cobra.Command, args []string) (error){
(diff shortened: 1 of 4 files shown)


jessealva:
@mdeuser ^^

mdeuser:
@jessealva - in general, when the pr is merged, no additional technical debt is expected.  in this case, let's bite the bullet and make the sweeping change prior to the pr merge.  for now, it's ok to do what you have suggested for quicker testing.

mdeuser:
@jessealva oh.. also, you will want to submit open cli PRs against the apache/incubator-openwhisk repo instead of this repo.  changes from the apache/incubator-openwhisk repo are then merged into this repo.

/cc @houshengbo

jessealva:
Closed as this was the incorrect repo in which to submit changes in.

PR #172, opened 2017-11-29
Change to ensure we can't pass in both --kind and --docker at the same time

### LINKED ISSUE #122: If you specify --kind and --docker, --docker is silently ignored
@sdague commented on [Thu Nov 16 2017](https://github.com/apache/incubator-openwhisk/issues/2977)

When you set an action using the wsk action update command, and specify both --kind and --docker, the --docker portion is silently ignored. Instead at least a warning should be provided to the users to ensure that they get something consistent.

This tripped me up when moving from a --kind python:3 to --docker <customimage> as I didn't remember to delete the --kind from my script, then my image wasn't used, the base python3action was used instead.

---

@csantanapr commented on [Thu Nov 16 2017](https://github.com/apache/incubator-openwhisk/issues/2977#issuecomment-345079610)

@jessealva take a look when you have a chance.

jessealva:
Fixes #122

### DIFF
--- a/commands/action.go
+++ b/commands/action.go
@@ -398,6 +398,11 @@ func parseAction(cmd *cobra.Command, args []string, update bool) (*whisk.Action,
         action.Annotations = annotations.(whisk.KeyValueArr)
     }
 
+    if len(Flags.action.kind) > 0 && len(Flags.action.docker) > 0 {
+        errStr := wski18n.T("Cannot specify both --kind and --docker at the same time.")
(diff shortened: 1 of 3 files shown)


jessealva:
@drcariel would you like to review this guy and see what you think?

jessealva:
@csantanapr could you review and merge this? I had @drcariel look at it but i guess he's not an approved reviewer so still have red checks on this PR.

PR #173, opened 2017-11-30
Added test cases for quoted strings.

jessealva:
These are some missing tests for incubator-openwhisk - PR 2462.  Not all strings are reachable...some are in debug messages and there is no facility for enabling debugging in the cli in tests right now. 

Also some of these are not reachable as there is no way for me to force a break of the code as the messages are embedded in an automated sdk install portion of the code.

I added as many tests as I could based on what was reachable via the testing framework.


jessealva:
@drcariel would you like to review this guy and see what you think?

tests/src/test/scala/system/basic/WskSdkTests.scala
```diff
@@ -82,6 +84,76 @@ class WskSdkTests extends TestHelpers with WskTestHelpers {
     }
   }
 
+  it should "download docker sdk when blackbox.tar.gz exists, and docker name should be quoted in error." in {
```
dubeejw: Could use `Seq` to squash this test and `"download ios sdk when OpenWhiskIOSStarterApp.zip exists, and ios name should be quoted in error.` into one test.

tests/src/test/scala/system/basic/WskSdkTests.scala
```diff
@@ -82,6 +84,76 @@ class WskSdkTests extends TestHelpers with WskTestHelpers {
...
+        workingDir = dir,
+        expectedExitCode = 1).stderr should include(
+        s"""The file '${fileName}' already exists.  Delete it and retry.""")
+    } finally {
+      file.delete()
+      FileUtils.deleteDirectory(dir)
+    }
+  }
+
+  it should "download ios sdk, check filename is quoted in error when create fails." in {
```
dubeejw: Can we expect the same directory permission issue with the Docker SDK?

```
$ mkdir wskinstall
$ chmod 555 wskinstall
$ cd wskinstall
$ wsk sdk install docker
error: The docker SDK installation failed: Error creating SDK file 'blackbox.tar.gz': open blackbox.tar.gz: permission denied
```

Could test both with `Seq`.

jessealva: I'll have to double check this one but last i checked i think there was no way to inject the bad permissions between the download and extraction.

houshengbo:
@jessealva please take a look at the issue: https://github.com/apache/incubator-openwhisk-cli/issues/221 Travis build is not happy at this moment, due to async with openwhisk.

rabbah:
@dubee this needs a rebase --- is this something you can do? otherwise going to close this.

PR #176, opened 2017-12-09
Only fetch action code when saving or copying an action

dubeejw:
Changes here prevent action code from being fetched unless an action is being saved. `wsk action get` will only fetch action code when `--save` or `--save-as` is specified. (Action code will still be fetched when copying an existing action.)

### DIFF
--- a/commands/action.go
+++ b/commands/action.go
@@ -35,27 +35,29 @@ import (
 )
 
 const (
-	MEMORY_LIMIT     = 256
-	TIMEOUT_LIMIT    = 60000
-	LOGSIZE_LIMIT    = 10
-	ACTIVATION_ID    = "activationId"
-	WEB_EXPORT_ANNOT = "web-export"
-	RAW_HTTP_ANNOT   = "raw-http"
-	FINAL_ANNOT      = "final"
-	NODE_JS_EXT      = ".js"
-	PYTHON_EXT       = ".py"
-	JAVA_EXT         = ".jar"
-	SWIFT_EXT        = ".swift"
-	ZIP_EXT          = ".zip"
-	PHP_EXT          = ".php"
-	NODE_JS          = "nodejs"
-	PYTHON           = "python"
-	JAVA             = "java"
-	SWIFT            = "swift"
-	PHP              = "php"
-	DEFAULT          = "default"
-	BLACKBOX         = "blackbox"
-	SEQUENCE         = "sequence"
+	MEMORY_LIMIT      = 256
+	TIMEOUT_LIMIT     = 60000
+	LOGSIZE_LIMIT     = 10
+	ACTIVATION_ID     = "activationId"
+	WEB_EXPORT_ANNOT  = "web-export"
+	RAW_HTTP_ANNOT    = "raw-http"
+	FINAL_ANNOT       = "final"
+	NODE_JS_EXT       = ".js"
+	PYTHON_EXT        = ".py"
+	JAVA_EXT          = ".jar"
+	SWIFT_EXT         = ".swift"
+	ZIP_EXT           = ".zip"
+	PHP_EXT           = ".php"
+	NODE_JS           = "nodejs"
+	PYTHON            = "python"
+	JAVA              = "java"
+	SWIFT             = "swift"
+	PHP               = "php"
+	DEFAULT           = "default"
+	BLACKBOX          = "blackbox"
+	SEQUENCE          = "sequence"
+	FETCH_CODE        = true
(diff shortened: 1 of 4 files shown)


dubeejw:
Example output: ```
./wsk action get test -i
ok: got action test
{
    "namespace": "test@test",
    "name": "test",
    "version": "0.0.1",
    "exec": {
        "kind": "nodejs:6"
    },
    "annotations": [
        {
            "key": "exec",
            "value": "nodejs:6"
        }
    ],
    "limits": {
        "timeout": 60000,
        "memory": 256,
        "logs": 10
    },
    "publish": false
}

```

**Comment by dubeejw** [comment]
@csantanapr, shall we provide an optional flag to display code from action get?

**Comment by rabbah** [comment]
We should. Did your pr for retrieving the code to file get migrated?

**Comment by csantanapr** [comment]
I was going to ask the same, what about —save that exports the code and just tell folks to uses that

**Comment by csantanapr** [comment]
If we have save and save-as I don’t think we need a flag

**Comment by dubeejw** [comment]
@mdeuser, any comments here?

**Comment by csantanapr** [comment]
@mdeuser bump

File: tests/src/test/scala/system/basic/WskBasicTests.scala [hunk 1]
```diff
@@ -313,7 +313,7 @@ class WskBasicTests extends TestHelpers with WskTestHelpers {
       wsk.action.get(name, fieldFilter = Some("exec")).stdout should include(s"""$successMsg""")
       wsk.action
         .get(name, fieldFilter = Some("exec"))
-        .stdout should include regex (s"""$successMsg exec\n\\{\\s+"kind":\\s+"nodejs:6",\\s+"code":\\s+"\\/\\*\\*[\\\\r]*\\\\n \\* Hello, world.[\\\\r]*\\\\n \\*\\/[\\\\r]*\\\\nfunction main\\(params\\) \\{[\\\\r]*\\\\n    greeting \\= 'hello, ' \\+ params.payload \\+ '!'[\\\\r]*\\\\n    console.log\\(greeting\\);[\\\\r]*\\\\n    return \\{payload: greeting\\}[\\\\r]*\\\\n\\}""")
+        .stdout should include regex (s"""$successMsg exec\n\\{\\s+"kind":\\s+"nodejs:6",\\s+"binary":\\s+false\\s+\\}""")
```
mdeuser: is the `binary` field new or just showing up?

tests/src/test/scala/system/basic/WskActionTests.scala
```diff
@@ -230,8 +230,6 @@ abstract class WskActionTests extends TestHelpers with WskTestHelpers with JsHel
     assetHelper.withCleaner(wsk.action, name) { (action, _) =>
       action.create(name, Some(TestCLIUtils.getTestActionFilename("empty.js")))
     }
-    val rr = wsk.action.get(name)
-    wsk.parseJsonString(rr.stdout).getFieldPath("exec", "code") shouldBe Some(JsString(""))
```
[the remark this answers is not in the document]
dubeejw: Validation happens above these lines during the create as a 200 status code is returned. Also three is a REST test here https://github.com/apache/incubator-openwhisk/blob/master/tests/src/test/scala/system/basic/WskActionTests.scala#L227.

tests/src/test/scala/system/basic/WskBasicTests.scala
dubeejw: It's new to the CLI.

tests/src/test/scala/system/basic/WskActionTests.scala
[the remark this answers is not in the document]
mdeuser: in theory, these base tests are common between the cli and REST test contexts.  removing these two lines seems to cause a test coverage discrepancy.  curious.. why are these two lines being removed in this repo, but not in the incubator-openwhisk repo?

commands/action.go
```diff
@@ -35,27 +35,29 @@ import (
...
+	NODE_JS           = "nodejs"
+	PYTHON            = "python"
+	JAVA              = "java"
+	SWIFT             = "swift"
+	PHP               = "php"
+	DEFAULT           = "default"
+	BLACKBOX          = "blackbox"
+	SEQUENCE          = "sequence"
+	FETCH_CODE        = true
+	DO_NOT_FETCH_CODE = false
```
mdeuser: perhaps a separate issue/pr.. but maybe some of these constants in this block are best defined in the incubator-openwhisk-client-go repo since they are more tightly coupled with api(s) defined in that repo.

PR #177, opened 2017-12-12
Move core/routemgmt & routemgmt install scripts to cli from main repo

ServoKvd:


rabbah:
The actions really belong with the API gateway.

PR #182, opened 2017-12-20
Change to use dep tool from golang

mcdan:


houshengbo:
@mcdan First, thanks for contributing to OpenWhisk CLI.
I believe there is something I need to explain a bit regarding the tests in CLI. By far, we have three type of tests available: unit tests(// +build unit), native tests(// +build native) and integration tests(// +build integration), with the marker on top of each test file.
Unit test is easy to understand. Native test means it is only able to run after the CLI binary is built and verify the functionalities of the binary without openwhisk services. Integration test means it is only able to run against the openwhisk services after the binary is built and verify the output of the openwhisk services.

IMO, unit tests definitely need to be put next to the file of source code, but for native tests and integration tests, I tend to put them in separate directories for better references. 

When travis launches, we run the unit tests first to verify the source code, then after we build the binary, we run the native tests to check whether the binary is built with the expected behaviors. Finally, we deploy openwhisk service, and run the integration tests to make sure we receive the expected results from openwhisk and process them correctly in CLI binary.

rabbah:
Closing as rotted and replaced by govendor.

PR #184, opened 2017-12-30
Create tarball for Nginx container in OpenWhisk as part of release deploy script

jonpspri:
In the interests of further removing circular dependencies, give ownership of `content.json` and a tarball for Nginx to this repository.  When this is done, everthing needed by OpenWhisk nginx to copy all CLI binaries to support local (inside-the-firewall) downloads of the CLI will be present in a single tarball.  What that means, in turn, is that the 'incubator-openwhisk' ansible scripts will no longer have to keep a list of supported architectures, and architectures can be added to the CLI simply by updating `content.json` in this archive.

This change should be backward compatible to the existing implementation and is a next step in decoupling the build architectues.  @rabbah @csantanapr @dgrove-oss @houshengbo 

Here's an example of the new tarball that is created (`OpenWhisk_CLI-latest-all_tree.tgz`) and would be added to the release page:

```
jonpspri@jyn:~/Projects/incubator-openwhisk-cli/build$ tar zftv OpenWhisk_CLI-latest-all_tree.tgz 
-rw-rw-r-- jonpspri/jonpspri 899 2017-12-30 08:50 content.json
-rw-rw-r-- jonpspri/jonpspri 3420521 2017-12-30 12:04 linux/386/OpenWhisk_CLI-linux-32bit.tgz
-rw-rw-r-- jonpspri/jonpspri 3681565 2017-12-30 12:04 linux/amd64/OpenWhisk_CLI-linux.tgz
-rw-rw-r-- jonpspri/jonpspri 3619707 2017-12-30 12:04 linux/s390x/OpenWhisk_CLI-linux-s390x.tgz
-rw-rw-r-- jonpspri/jonpspri 3413129 2017-12-30 12:05 mac/386/OpenWhisk_CLI-mac-32bit.zip
-rw-rw-r-- jonpspri/jonpspri 3643829 2017-12-30 12:05 mac/amd64/OpenWhisk_CLI-mac.zip
-rw-rw-r-- jonpspri/jonpspri 3366514 2017-12-30 12:05 windows/386/OpenWhisk_CLI-windows-32bit.zip
-rw-rw-r-- jonpspri/jonpspri 3607470 2017-12-30 12:06 windows/amd64/OpenWhisk_CLI-windows.zip
```

The content.json file used to generate this is at https://github.com/jonpspri/incubator-openwhisk-cli/blob/nginx-aware-build/config/content.json.


jonpspri:
Added POWER (ppc64le) and Pi (arm and arm64) architectures to the build.  Discovered a GO-ism that needed to be addressed in the Travis configuration as a result.

config/content.json
```diff
@@ -0,0 +1,49 @@
+{
```
dubeejw: @jonpspri, is it possible to automatically generate content.json in the same fashion that build.sh (https://github.com/apache/incubator-openwhisk-cli/blob/master/build.sh) does?

jonpspri: Yes, I think we can, but I'm not sure we want to.  The key configuration attributes are:
1.  OS Architecture combination,
2.  Filename for the tar file (with some interesting exception cases for naming standards).

I like having 'content.json' driving the build, because configuration directly drives content.  We _could_ add one layer of indirection so we don't manually maintain the directory structure and file names, but I'm not sure it's worth the work given that this configuration seems unlikely to change frequently.  Thoughts from any/all?

I will look at build.sh and also at what's over in incubator-openwhisk.  I welcome further input from you and others, especially on cases for where the Filename may need substitution.

dubeejw: We can generate content.json from a few simple configuration arguments. For instance, build.gradle tells build.sh which OSes and architectures to build for. See https://github.com/apache/incubator-openwhisk-cli/blob/master/build.gradle#L57.

dubeejw: The above process can be made more sophisticated in order to account for the additional linux architectures.

jonpspri: Okay, but what's the future of the Gradle-based build?  From the README.md it looked like it was to be depricated in favor of a straight Go build.  And the Travis build for deployment runs the go builds directly in the Travis environment.  My thinking when I wrote this was that config.json was an already-defined container for the 'few simple configuration arguments' you mention above; however, if we're going to support gradle builds as well then yes, we can work out how to integrate the processes.|

I sincerely apologize if I'm coming off as argumentative -- that's not my intent.  I just want to make an informed decision about how to balance the various needs of the build (and consumption) processes.  So give me some time to digest what's going on with the Gradle/Docker build.

jonpspri: @dubeejw Could you take a look at https://github.com/jonpspri/incubator-openwhisk/blob/2f90600617803cff7c4cae7d8880c1cad8e12160/ansible/roles/cli/files/createContentJson.sh ?  I essentially did there what would need to happen to build.sh in order to generate content.json .  In addtion, the arguments format for the docker build would need to change.  CLI_OS and CLI_ARCH as separate variables no longer works because once you introduce the more esoteric architectures like s390x and ppc64le, the build list is no longer a cartesian product of the two lists.  I've had to resort to lists of pairs instead to make the build work.  The reason I wasn't generating content.json is because it essentially IS that list of pairs.

dubeejw: I'll have to take a closer look when I get a chacne. All input/conversation is welcome, argumentative or not. :)

jonpspri: Meanwhile, as I've been poking around I've arrived at the conviction that the real problem is that we've got builds happening in two places:  'build.sh' (via gradle) and 'tools/travis/build_tag_releases.sh'.  So, I'm going to tinker with how the travis build works to see if I can get it to use 'build.sh'.  That will mean figuring out how to spec a cross-compile build via ./gradlew, though, which may bring us back around to the 'config.json' conversation.

dubeejw: @houshengbo, can you provide any info regarding the above?

dubeejw: I think we are fine with getting rid of the Docker build parts in favor of a straight Go build. The Docker build was previously used when the CLI resided in the main OpenWhisk repo. 

Perhaps we can use Gradle to kick off the Go build instead of Make.

jonpspri: Okay, I'm going to start unwinding builds with a goal of having build logic reside in only one place and have something (Gradle or Make tbd) drive the build.  I'm up in the air about which build tool to use -- Gradle is great for the JVM-based languages with CLASSPATHs, etc, but Make may be a better answer for go, since it's more compiler-like.  We'll burn that bridge when we come to it -- first goal for me is eliminating the need for 'tools/travis/build_tag_releases.sh'.

houshengbo:
@jonpspri If I understand it correctly, we are about to use content.json to save the supported architectures and file names, in order to run the build to generate the binaries. As long as it is defined in content.json, the binary is generated. The file content.json becomes the input of the binary build. 

One of my concerns is that what are the default binaries to be generated if content.json is missing or empty?

Another remark is to make sure if does not break the gradle build, because we support to build the binaries by ./gradlew buildBinaries -PcrossCompileCLI=true. If we build all the binaries, will it pick up the input from content.json?

Third, how do the user customize which binaries they need? I hope we can generate content.json with different input of architectures and file name.

At least, we need to have the solutions to the above question marks.

jonpspri:
Good question marks.  I'm going to come at them sideways first, because I want to be sure there is clarity in the discussion.

TL;DR:  I hadn't intended to work on the gradle build because it isn't used to build the release files.  But probably all the build processes should be the same, and I'm open to suggestions on what that unified build should look like (though I'm leaning toward a Makefile).

Context:  I started down this path because I was introducing OpenWhisk to the Linux on Z ('linux s390x') architecture.  The immediate problem I ran into was that 's390x' wasn't a valid architecture for mac/darwin or windows OSs, which broke the cartesian logic that had been used to build the CLI.  I made some corrections in the 'incubator-openwhisk' and 'incubator-openwhisk-cli' to address that.

What became clear in the longer term is that process and control were flowing back and forth between the two projects.  To add an os/architecture required 1) Adding it to the CLI compile and production repository, 2) Adding it to the list of os/architectures to be downloaded/redirected to, and 3) generating a 'content.json' file as part of the ansible deployment scripts for OpenWhisk to make known what was available from a particular OpenWhisk build.  Which led me to my axiom for this PR:

**Axiom:**  The 'incubator-openwhisk-cli' project should have the master list of what CLI builds are available and package them along with content.json.  Then the CLI process becomes one-way -- the ansible Nginx deployment process in 'incubator-openwhisk' downloads a tarball and we're off to the races.

So I did what I'd been doing, I fiddled with the Travis scripts to build an extra artifact for release -- a tarball of content.json and the various builds of wsk to be downloaded directly to Nginx.  As I was thinking this through I came to a conclusion:  

**Conclusion:*** "Hey content.json is metadata that could be produced manually and used to drive the build".  Note, that it doesn't have to be this way, but the list of os/architecture pairs has to live somewhere -- if not in content.json then in some other build file or configuration file.  I grant hand-coding JSON is not the most romantic endeavor, but it may be less resource intensive than maintaining something fancier.

Okay, so here I was with a beautiful PR that would make what I wanted when my new friend @dubeejw pointed out "Hey what about the Gradle build?", to which my initial response was "WTF -- Isn't the gradle build being discouraged anyway?".  On re-reading though, I realized it wasn't and that even my 's390x' patch hadn't made it to that, which gets us to where we are.

So: what are the eventual goals here?  This is what I think we want:
1.  Travis builds and releases each individual binary.
2.  Travis builds and releases a tarball suitable for Nginx to download (including content.json)
3.  A developer can build the local os/architecture combination
4.  A developer can build a _single_ specified os/architecture combination
5.  A developer can build all support cross-compile combinations (and get a bonus content.json)

And ideally, we would do this:
1.  With a minimum of duplicated code between the travis and developer approaches, and
2.  Using a local go installation (instead of resorting to Docker)

My gut says the way to achieve this is with gradle/make and the existing build.sh script, though I may move some logic to and fro between.  Travis would invoke gradle/Make.  We'd rip out all the gradle and docker stuff, but the build would target a '.gitignore'd build directory.

I'm open to opinions about build tool.  I think Make is, in the longer run, a lighter-weight very capable choice, but if we think openwhisk devs are more likely to talk Gradle, we can go that way.

So to answer your questions (and this is vision, not how the code looks today):
1.  Configuration is part of the repository/source code and so should not be empty.  Build options are current architecture (go without options), specified architecture (GOOS/GOARCH) or cross-compile everything.
2.  Cross-compile build becomes a target in the overall build script.
3.  Architecture(s) to build are a build script parameter.

I think the key point is to make sure we reduce the processing flow to a single flow, which is where I want to focus for now.  I will let you know whether that becomes another PR.

Sorry for the _very_ long reply, but it's as much to get my thinking clear as anything.  Kudos if you made it this far!

jonpspri:
@houshengbo @dubeejw Hi gents.  I'm obviously still running into some Travis build challenges, but I wanted you to see what I've been working on.  As of these commits, all automated builds _should_ be able to process via Gradle build.  In addition, the [gogradle](https://github.com/gogradle/gogradle) plugin that I used obviates the need for Docker and handles the Go and dependency configuration for us.  Net-net, the build.gradle is a bit more compilicated, but it does what had previously been accomplished by Docker/Travis and the build*.sh scripts.

I'm considering what to do with the Makefile, which still has some redundancies to the Gradle build and working to test cleanly now, but I wan't you to have the chance to comment on this approach.

jonpspri:
@dubeejw @houshengbo Hi gents.  I think this is ready for review and comment.  I expect we will go through a round of suggested revisions, so please let me know what changes you may want to see.  OTOH, I'd like to clear this from my plate, so please start reviewing ASAP.  Thanks, -J

dubeejw:
@jonpspri, I will get to the review of this today, thanks.

dubeejw:
Can you rebase to fix the conflict with build_tag_releases.sh?

.travis.yml
```diff
@@ -8,28 +8,40 @@ matrix:
...
 
 install:
-  - ./tools/travis/setupscan.sh
   - export DEPLOY_BUILD_READY=false
-  - go get -u github.com/golang/lint/golint
-  - go get -u github.com/stretchr/testify
-  - go get -u github.com/spf13/viper
+
+before_script:
+  # Clone needed git repositories for the build -- TODO: Pick a commit?
```
dubeejw: I think running against master is fine.

.travis.yml
```diff
@@ -8,28 +8,40 @@ matrix:
...
+  - cd gogradle && ./gradlew --console=plain assemble
 
 script:
+  - cd $TRAVIS_BUILD_DIR
+  - ./gradlew --console=plain buildBinaries release -PcrossCompile=true -PzipFileName=${zip_file_name}
   - ./tools/travis/test_openwhisk.sh
 
-after_script:
-  - make clean
+#after_script:
```
dubeejw: Can these commented lines be removed?

.travis.yml
```diff
@@ -44,24 +56,19 @@ after_success:
...
 before_deploy:
-  - go get github.com/inconshreveable/mousetrap
-  - go get github.com/mattn/go-isatty
-  - export build_file_name=wsk
-  - export zip_file_name=OpenWhisk_CLI
-  - chmod +x tools/travis/build_tag_releases.sh
-  - ./tools/travis/build_tag_releases.sh $build_file_name $zip_file_name
-  - export RELEASE_PKG_FILE=$(ls $zip_file_name-*.*)
-  - echo "Deploying $RELEASE_PKG_FILE to GitHub releases."
+  - echo "Deploying $(ls ./release/${zip_file_name}-*.tgz ./release/${zip_file_name}-*.zip) to GitHub releases."
```
dubeejw: `echo "Deploying $(ls ./release/${zip_file_name}-*.tgz, and ./release/${zip_file_name}-*.zip) to GitHub releases."`

README.md
```diff
@@ -4,65 +4,85 @@
...
+The OpenWhisk CLI is available on the release page: [click here to
+download](https://github.com/apache/incubator-openwhisk-cli/releases). We
+currently have binaries available for Linux, Mac OS and windows under amd64
+architecture. You can download the binary, which fits your local environment.
 
 
 # How to build the binary locally
 
-You can also choose to build the binaries locally from the source code with Go tool.
+The OpenWhisk CLI is written in the Go language.  You have two options to build
```
dubeejw: Remove double space before second sentence for consistency.

README.md
```diff
@@ -4,65 +4,85 @@
...
+the binary locally:
+
+1.  Compile in your logal Go environment,
```
dubeejw: `local` instead of `logal`.

README.md
```diff
@@ -4,65 +4,85 @@
...
 ```
 $ go build -o wsk
 ```
 
-If you would like to build the binary for a specific operating system, you may add the arguments GOOS and
-GOARCH into the Go build command. Since it is only applicable under amd64 architecture, you have to set GOARCH
-to amd64. GOOS can be set to "linux" "darwin" or "windows".
+If you would like to build the binary for a specific operating system, you may
+add the arguments GOOS and GOARCH into the Go build command. Since it is only
+applicable under amd64 architecture, you have to set GOARCH to amd64. GOOS can
```
dubeejw: Is this right, can't we build for other architectures besides amd64?

README.md
```diff
@@ -4,65 +4,85 @@
...
 
-## Build the binary with Docker and Gradle
+## Build the binary with Gradle
 
-This is the second choice for you to build the binary. Make sure that you have Docker and gradle on your machine:
-[installing Docker](https://docs.docker.com/engine/installation/) and [installing Gradle](https://gradle.org/install) for your local machine.
+**Note:** For those who may have used the Gradle build previously, it has been
+re-engineered to no longer required Docker or Go to be pre-installed on your
+system.  Using the [gogradle](https://github.com/gogradle/gogradle) plugin, the
+script now uses a go environment if you have it, but will also create a local
```
dubeejw: `Go` instead of `go`.

README.md
```diff
@@ -4,65 +4,85 @@
...
-If it is executed successfully, you can find your binary `wsk` directly under OpenWhisk CLI home directory.
+If it is executed successfully, you can find your binary `wsk` directly under
+OpenWhisk CLI home directory.
 
-## Build the binary with Docker and Gradle
+## Build the binary with Gradle
 
-This is the second choice for you to build the binary. Make sure that you have Docker and gradle on your machine:
-[installing Docker](https://docs.docker.com/engine/installation/) and [installing Gradle](https://gradle.org/install) for your local machine.
+**Note:** For those who may have used the Gradle build previously, it has been
```
dubeejw: Yay @ Gradle without Docker!

README.md
```diff
@@ -4,65 +4,85 @@
...
+re-engineered to no longer required Docker or Go to be pre-installed on your
+system.  Using the [gogradle](https://github.com/gogradle/gogradle) plugin, the
+script now uses a go environment if you have it, but will also create a local
+Go environment if you do now.
```
dubeejw: `if one already exists`.

README.md
```diff
@@ -4,65 +4,85 @@
...
 
-After that, open an terminal, go to the directory of OpenWhisk CLI home directory, and
-build the binary via the following command under Linux or Mac:
+To build with Gradle, open an terminal, go to the directory of OpenWhisk CLI
+home directory, and build the binary via the following command under Linux or
+Mac:
```
dubeejw: Can we add Windows to this now?

README.md
```diff
@@ -74,40 +94,56 @@ or run the following command for Windows:
...
 
-If you would like to build the binaries available for all the operating systems and architectures, run the following
-command:
+
+If you would like to build the binaries available for all the operating systems
+and architectures, run the following command:
 
 ```
-$ ./gradlew buildBinaries -PcrossCompileCLI=true
+$ ./gradlew buildBinaries -PcrossCompile=true
```
dubeejw: Reducing redundancy here, good. 👍

README.md
```diff
@@ -74,40 +94,56 @@ or run the following command for Windows:
...
 ```
 
-Then, you will find the binaries and their compressed packages generated under the folder bin/\<os\>/\<cpu arc\>/ for each
-operating system and CPU architecture pair. We supports both amd64 and 386 for Linux, Mac and Windows operating systems.
+Then, you will find the binaries and their compressed packages generated under
```
dubeejw: `Then you` instead.

README.md
```diff
@@ -74,40 +94,56 @@ or run the following command for Windows:
...
+the folder `bin/<os>/<cpu arc>/` for each operating system and CPU
+architecture pair. We supports both amd64 and 386 for Linux, Mac and Windows
```
dubeejw: `We support` instead of `We supports`. Is the list of architectures correct?

README.md
```diff
@@ -74,40 +94,56 @@ or run the following command for Windows:
...
+operating systems.
+
+To specify a build for specific architectures, you can provide a comma- or
```
dubeejw: `comma` instead of `comma-`.

build.gradle
```diff
@@ -1,93 +1,286 @@
-ext.dockerImageName = "cli"
-ext.dockerContainerName = "cli"
-ext.dockerBuildArgs = getDockerBuildArgs()
-apply from: 'gradle/docker.gradle'
+// plugins {
```
dubeejw: Remove commented lines here.

build.gradle
```diff
@@ -1,93 +1,286 @@
...
-                ":/src/github.com/apache/incubator-openwhisk-cli/build/.", "${projectDir}/bin"])
+golang {
+    packagePath = 'github.com/apache/incubator-openwhisk-cli' as String
+    buildTags = (rootProject.findProperty('goTags')?:'').split(',')
+}
 
-        run(dockerBinary + ["rm", "-f", dockerContainerName])
+dependencies {
+    golang {
+        // TODO - validate where version should and should not be locked.
```
dubeejw: Versions need be locked for every imported package for legal reasons and to prevent random breakages. How does this affect Godeps? Do we need to lock the commit hashes in two places now? Will running `go get` not pick these dependancies up?

build.gradle
```diff
@@ -1,93 +1,286 @@
...
-        println "go.arch = "+mapOsArchToGoArch(getOsArch())
+/*
+    The OpenWhiskPlatform class is a utility class to make the rest of what
+    happens with platforms a bit more understandable.  A "Platform" is a tuple
+    of an operating system and a processor.  Currently, the OpenWhisk CLI
+    supports three OS's:  Linux, Mac/Darwin, and Windows.  It supports x86
+    (32-bit or 64-bit) on all OS's.  On Linux, it also support System Z (s390x),
+    PowerPC (ppc64le), and ARM (32-bit and 64-bit) architectures.
+
+    Different contexts us different codings to refer to these architectures --
```
dubeejw: `use` instead of `us`.

build.gradle
```diff
@@ -1,93 +1,286 @@
...
+def content = [:]
+for (p in platforms) {
+    def pathObject = [ "path" : "${p.archiveDirName}/${p.archiveFileName}" ]
+    content.get(p.owOs,[:])[p.goArch] = pathObject
+    if (p.goArch == 'amd64') { content.get(p.owOs,[:])['default'] = pathObject }
 }
 
-def getOsArch() {
-    return System.properties['os.arch']
+task contentJson() {
```
dubeejw: 👍

tools/travis/test_openwhisk.sh
```diff
@@ -2,57 +2,87 @@
...
-# Build script for Travis-CI.
-WHISKDIR="$HOMEDIR/incubator-openwhisk"
+#  Run separate test scopes as separate
+./gradlew --console=plain goLint
+./gradlew --console=plain goTest -PgoTags=unit
+export PATH=$PATH:$TRAVIS_BUILD_DIR
+./gradlew --console=plain goTest -PgoTags=native
 
-cd $WHISKDIR
+#
```
dubeejw: Remove this line.

tools/travis/test_openwhisk.sh
```diff
@@ -2,57 +2,87 @@
...
+#  Set up the OpenWhisk environment ( TODO: reusable script for incubtor-openwhisk? )
```
dubeejw: Not sure what this comment means.

tools/travis/test_openwhisk.sh
```diff
@@ -2,57 +2,87 @@
...
 $ANSIBLE_CMD couchdb.yml
 $ANSIBLE_CMD initdb.yml
 $ANSIBLE_CMD apigateway.yml
-
-cd $TRAVIS_BUILD_DIR
-TERM=dumb ./gradlew buildBinaries
-
-cd $WHISKDIR/ansible
 $ANSIBLE_CMD wipe.yml
+# TODO -- Some flag might be needed to get CLI from local directory (?)
```
dubeejw: Nope. What you have is right.

Dockerfile
```diff
@@ -1,27 +0,0 @@
-FROM golang:1.8
```
dubeejw: My first Dockerfile. Bye 👋

README.md
dubeejw: Line 88 and 94 should be `distDocker` instead of `buildBinaries`.

gradle/docker.gradle
```diff
@@ -3,7 +3,7 @@ import groovy.time.*
 /**
  * Utility to build docker images based in gradle projects
  *
- * This extends gradle's 'application' plugin logic with a 'buildBinaries' task which builds
```
dubeejw: Can this file be deleted since the Dockerfile was removed?

tools/travis/test_openwhisk.sh
```diff
@@ -2,57 +2,87 @@
...
 
-cd $WHISKDIR
+#
+#  Set up the OpenWhisk environment ( TODO: reusable script for incubtor-openwhisk? )
+cd $OPENWHISK_HOME
 ./tools/travis/setup.sh
 
 ANSIBLE_CMD="ansible-playbook -i environments/local -e docker_image_prefix=testing"
-TERM=dumb ./gradlew distDocker -PdockerImagePrefix=testing
+./gradlew --console=plain distDocker -PdockerImagePrefix=testing
```
dubeejw: Should this be `buildBinaries` instead of `distDocker`?

README.md
```diff
@@ -4,65 +4,85 @@
...
+the command line and automate them through scripts.
 
 # Where to download the binary of OpenWhisk CLI
 
-The OpenWhisk CLI is available on the release page: [click here to download](https://github.com/apache/incubator-openwhisk-cli/releases).
-We currently have binaries available for Linux, Mac OS and windows under amd64 architecture. You can download the
-binary, which fits your local environment.
+The OpenWhisk CLI is available on the release page: [click here to
+download](https://github.com/apache/incubator-openwhisk-cli/releases). We
+currently have binaries available for Linux, Mac OS and windows under amd64
```
dubeejw: I think this incorrect since more architectures are added.

tools/travis/test_openwhisk.sh
dubeejw: Why do we need to build again here, shouldn't the binaries already exist from the build?

.travis.yml
```diff
@@ -8,28 +8,40 @@ matrix:
...
+  - git clone --depth 3 https://github.com/apache/incubator-openwhisk.git
+  - export OPENWHISK_HOME="$(cd '../incubator-openwhisk' && pwd )"
+  - git clone --depth 3 https://github.com/apache/incubator-openwhisk-utilities.git
+  # These should be temporary until next release of gogradle
+  - git clone --depth 3 https://github.com/jonpspri/gogradle.git
+  - cd gogradle && ./gradlew --console=plain assemble
 
 script:
+  - cd $TRAVIS_BUILD_DIR
+  - ./gradlew --console=plain buildBinaries release -PcrossCompile=true -PzipFileName=${zip_file_name}
```
dubeejw: test_openwhisk.sh builds again, is it needed there?

.travis.yml
jonpspri: Okay.  TODO removed.

.travis.yml
jonpspri: Yes.  Done.

.travis.yml
jonpspri: Not going to give the result you expect.  The shell expansion is creating a list of all filenames so the , and is out of place.  Let's see if this format is intolerable -- we can get more clever if it is.

README.md
jonpspri: Good point.  Re-written to include other architectures.

README.md
jonpspri: Oh, c'mon.  It doesn't show when it renders!  Done anyway :)

README.md
jonpspri: Fixed.

README.md
jonpspri: Good catch.  Deleted the 'specific architectures' sentence since it seems to function under any architecture Go supports.

README.md
jonpspri: D'oh (no d'oh).  Fixed.

README.md
jonpspri: Ugh this is a tough construction.  I tried a different rewrite -- see if it works for you.

README.md
jonpspri: Windows command is two lines down from this (.\gradlew.bat).

README.md
jonpspri: I'm lazy.  Less typing is also good.

README.md
jonpspri: Actually, 'The build script will place ' gives us a better idea of what's really happening.

README.md
jonpspri: Nope.  Re-written.

README.md
jonpspri: Grr.  Didn't refer to my style-book :)  Fixed.

build.gradle
jonpspri: Actually, would like to hold them as they'll be restored when gogradle 0.8.1 comes out.  Added a note about that.

build.gradle
jonpspri: The gogradle script supports locking.  I'll put it into place for the next build cycle.  You can see the commits already in place for most imports, grabbed from godeps.  Gogradle takes over management of the locks from gogradle.  I'll take a look at #201 see that everything's buttoned up.  Will also document locking in the script.

build.gradle
jonpspri: Corrected.

gradle/docker.gradle
jonpspri: Not yet, because it's still used to build the images for 'incubator-openwhisk' in support of the testing.  In time we can try to get the build to use the docker.gradle in 'incubator-openwhisk', but I'm trying to manage the number of moving parts.

tools/travis/test_openwhisk.sh
jonpspri: Done.

tools/travis/test_openwhisk.sh
jonpspri: Ah!  It's a note to myself that there's an issue with the broader test environment and strategy for all these sub-projects that depend on having an OpenWhisk environment running for integration testing.  I was thinking perhaps such an environment script should be part of the 'incubator-openwhisk' repository, but again -- too many moving parts.  I'll rewrite the comment for now.

tools/travis/test_openwhisk.sh
jonpspri: Check the working directory.  This is actually a build of OpenWhisk (not the CLI) for integration so the integration environment is running.  So this really is a build of the docker images OpenWhisk uses to run its cluster.

tools/travis/test_openwhisk.sh
jonpspri: Okay.  I'm taking your word for it for the moment, but when I go back to OpenWhisk to get the NGinx to download the tarball (where this all started), I'm going to take a second look.  Meanwhile, deleting the comment.

.travis.yml
jonpspri: I think the build was removed from test_openwhisk.sh (see my comments above).  In fact, this should be the only build that happens in the Travis processing.

jonpspri:
Addressed most of the comments -- I'm now re-basing and pushing.  

I need to take a look at the gogradle docs to lock the dependencies -- that's very doable.  As you can see, it can also be done transparently in the build.gradle file.

houshengbo:
@jonpspri please do a rebase of the latest code. The master is currently failing to build to the binaries with the 'go build' command due to the following errors:

```
commands/api.go:106: cannot range over api.Swagger.Paths[path] (type *whisk.ApiSwaggerPath)
commands/api.go:211: cannot range over retApi.Swagger.Paths[path] (type *whisk.ApiSwaggerPath)
commands/api.go:580: cannot range over resultApi.Swagger.Paths[path] (type *whisk.ApiSwaggerPath)
commands/api.go:618: cannot range over resultApi.Swagger.Paths[path] (type *whisk.ApiSwaggerPath)
commands/api.go:681: cannot range over resultApi.Swagger.Paths[path] (type *whisk.ApiSwaggerPath)
commands/api.go:715: cannot range over resultApi.Swagger.Paths[path] (type *whisk.ApiSwaggerPath)
```

jonpspri:
Huh.  I just did a rebase from Apache/...:master and force push with the last commit.  I will try to figure out what’s going on; I suspect dependency issues because they’re all Swagger-related errors.  I will try to get a clean build and reproduce the error you’re getting.

houshengbo:
@jonpspri I think @dubeejw had a PR as a quick fix: https://github.com/apache/incubator-openwhisk-cli/pull/201

It should work fine after fixing my comments.

jonpspri:
@houshengbo It looks like the root cause was that the `openwhisk-gradle-go-client` library moved underneath me and had been locked by the dependencies.  I added instructions on managing `vendor` to the README.md.  Net-net, run `./gradlew goVendor` to make sure `vendor` contains the appropriate libraries before build.  That essentially replaces godeps (per #201).  See if that works for you and thanks all for the documentation proofreading.

jonpspri:
Ah crud.  Looks like I need to rebase again...

dubeejw:
@jonpspri, looks like goLint failed in your Travis build.

Godeps/Godeps.json
```diff
@@ -1,71 +0,0 @@
-{
```
dubeejw: I think we need to keep this file to lock dependency versions into place.

build.gradle
dubeejw: We need to lock the dependancies using `godep` though.

build.gradle
```diff
@@ -1,93 +1,290 @@
...
+        build(['name':'github.com/mitchellh/go-homedir', 'version':'1111e456ffea841564ac0fa5f69c26ef44dafec9', 'transitive':false])
+        build(['name':'github.com/nicksnyder/go-i18n/i18n/...', 'version':'37e5c2de3e03e4b82693e3fcb4a6aa2cc4eb07e3', 'transitive':false])
+        build(['name':'github.com/spf13/cobra', 'version':'1238ba19d24b0b9ceee2094e1cb31947d45c3e86', 'transitive':false])
+        build(['name':'github.com/spf13/pflag', 'version':'367864438f1b1a3c7db4da06a2f55b144e6784e0', 'transitive':false])
+        build(['name':'golang.org/x/sys/unix', 'version':'7f918dd405547ecb864d14a8ecbbfe205b5f930f', 'transitive':false])
+        build(['name':'gopkg.in/yaml.v2', 'version':'cd8b52f8269e0feb286dfeef29f8fe4d5b397e0b', 'transitive':false])
+        build(['name':'github.com/ghodss/yaml', 'version':'0ca9ea5df5451ffdf184b4428c902747c2c11cd7', 'transitive':false])
+        build(['name':'github.com/apache/incubator-openwhisk-client-go/whisk','version':'a67e8509a92beb6c68f0c9da43562af1f5d2b13c','transitive':false])
+        // END - Imported from Godeps
+        test name:'github.com/golang/lint/golint', version:'e14d9b0f1d332b1420c1ffa32562ad2dc84d645d', transitive:false
```
dubeejw: The linter needs to be a build dependency instead of a test dependency.

dubeejw:
@jonpspri, here's some changes I had in mind https://github.com/jonpspri/incubator-openwhisk-cli/pull/1.

build.gradle
jonpspri: Actually, the linter needs to be pulled in in travis.yml, because the build expects the binary to be in the execution path.  I made that change to travis.yml.

jonpspri:
I looked at jonpspri#1 and integrated what I could.  My PR for gogradle didn't make it into 0.8.1, so I still have to grab that from my fork for the foreseeable future.  I did most of the refactoring you proposed, except I don't include clean in the dependencies for anything (1) because it would wipe out content.json just before we need it for the archive, and (2) at least in my old *ix experience, clean was always invoked by the user from the command line.  Also, I left the creation of './bin/wsk' in because it's needed by the testing and just generally nice to have.  I added comments because it wasn't obvious from the code what was going on.

dubeejw:
Restarted Travis.

tools/travis/test_openwhisk.sh
```diff
@@ -2,57 +2,85 @@
...
-export PATH=$PATH:$TRAVIS_BUILD_DIR;
-make native_test;
-
-export OPENWHISK_HOME="$(dirname "$TRAVIS_BUILD_DIR")/incubator-openwhisk";
+#
+#  At this point, the Travis build should already have built the binaries and
+#  the release.  If you're running manually, this command should get you to
+#  the same place:
+#
+#    ./gradlew buildBinaries release
```
dubeejw: This should be `./gradlew release`.

build.gradle
```diff
@@ -1,93 +1,284 @@
...
 }
 
-def run(cmd, ignoreError = false) {
-    println("Executing '${cmd.join(" ")}'")
-    def proc = cmd.execute()
-    proc.waitFor()
-    if(!ignoreError && proc.exitValue() != 0) {
-        println("Command '${cmd.join(" ")}' failed with exitCode ${proc.exitValue()}")
+task compile(type: Copy, dependsOn: goBuild) {
+    destinationDir = file('./bin')
```
dubeejw: Do we really need the bin folder, or can we just rely on the build folder?

build.gradle
```diff
@@ -1,93 +1,284 @@
...
-    String osname_l = osname.toLowerCase()
-    if (osname_l.contains("nux") || osname.contains("nix")) return "linux"
-    if (osname_l.contains("mac")) return "mac"
-    if (osname_l.contains("windows")) return "windows"
-    return osname_l
+task release(type: Tar, dependsOn: [individualArchives, index]) {
+    compression = Compression.GZIP
+    destinationDir = file('./release')
+    baseName = "${OpenWhiskPlatform.zipFileName}-latest-all"
+    from('./release/content.json') { into('.') }
```
dubeejw: `content.json` is now lives in the build folder.

build.gradle
dubeejw: We would just need to copy the native binary to `build/wsk` here for use with the tests.

tools/travis/test_openwhisk.sh
```diff
@@ -2,57 +2,85 @@
...
-cd $WHISKDIR/ansible
 $ANSIBLE_CMD wipe.yml
 $ANSIBLE_CMD openwhisk.yml -e openwhisk_cli_home=$TRAVIS_BUILD_DIR
 
 # Copy the binary generated into the OPENWHISK_HOME/bin, so that the test cases will run based on it.
-mkdir -p $WHISKDIR/bin
-cp $TRAVIS_BUILD_DIR/bin/wsk $WHISKDIR/bin
+# TODO - if the ansible above were correctly configured, this wouldn't be necessary
+mkdir -p $OPENWHISK_HOME/bin
+cp -f $TRAVIS_BUILD_DIR/bin/wsk $OPENWHISK_HOME/bin
```
dubeejw: Coulld be `build/wsk`.

build.gradle
jonpspri: I like the idea of using the build folder -- one less mouth to feed.  I'll make the adjustment(s).

build.gradle
jonpspri: Fixed.

tools/travis/test_openwhisk.sh
jonpspri: Agreed.  Will work it through.

jonpspri:
Latest commit adds an automatic 'go get' of the golint binary and removes the need for a bin directory.  Getting closer...

README.md
```diff
@@ -4,110 +4,167 @@
...
+the command line and automate them through scripts.
 
-The OpenWhisk Command Line Interface (OpenWhisk CLI) is a unified tool that provides a consistent interface to
-interact with OpenWhisk services. With this tool to download and configure, you are able to manage OpenWhisk services
-from the command line and automate them through scripts.
+# Where to download the binary of OpenWhisk CLI
 
+The OpenWhisk CLI is available on the release page: [click here to
+download](https://github.com/apache/incubator-openwhisk-cli/releases). We
+currently have binaries available for Linux, Mac OS and windows under i386 and
```
dubeejw: `Windows` instead of `windows`.

README.md
```diff
@@ -4,110 +4,167 @@
...
+or run the following command for Windows:
 
 ```
-$ cd $GOPATH
-$ go get github.com/apache/incubator-openwhisk-cli
+$ ./gradlew.bat compile -PnativeCompile
 ```
 
+After the build, you can find the binary `wsk` or `wsk.exe` in the bin folder
+under the OpenWhisk CLI home directory. In addition, it is also available under
```
dubeejw: Need to remove these two sentences that reference the `bin` folder.

README.md
dubeejw: Also a few sentences in here that start with double spaces. I know, I know.

README.md
```diff
@@ -4,110 +4,167 @@
...
+the folder `bin/<os>/<architecture>/`. For example, if your local operating
+system is Mac, and the CPU architecture is amd64, the binary and its compressed
+package can also be found under `bin/mac/amd64/`.
 
-## Build the binary with Go
+OpenWhisk CLI(`wsk`) is produced in a Docker container during the build process
```
dubeejw: I don't think this paragraph is true anymore.

gradle/docker.gradle
dubeejw: From test_openwhisk.sh:

```
+cd $OPENWHISK_HOME/ansible
 +ANSIBLE_CMD="ansible-playbook -i environments/local -e docker_image_prefix=testing"
```

Doesn't the above mean the docker.gradle file from the OpenWhisk directory is being used?

README.md
jonpspri: Fixed.

README.md
jonpspri: Re-written.

README.md
jonpspri: Agreed.  Deleted the whole paragraph.

jonpspri:
Regarding the need for 'docker.gradle', the tests that break without it are toward the bottom of the test_openwhisk.sh:

```
#  Run the test cases under openwhisk to ensure the quality of the runnint API.
cd $TRAVIS_BUILD_DIR
./gradlew --console=plain :tests:test -Dtest.single=*ApiGwCliTests*
sleep 30
./gradlew --console=plain :tests:test -Dtest.single=*ApiGwCliRoutemgmtActionTests*
sleep 30
./gradlew --console=plain :tests:test -Dtest.single=*ApiGwCliEndToEndTests*
sleep 30
./gradlew --console=plain :tests:test -Dtest.single=*Wsk*Tests*
```

jonpspri:
@dubeejw But as I walk back from getting water, I'm wondering whether we could/should run some/all of those same tests from within $OPENWHISK_HOME instead.  We might be able to do an end run on need for docker.gradle.

jonpspri:
The dependency comes from `tests/build.gradle`:

```
dependencies {
    compile "org.scala-lang:scala-library:${gradle.scala.version}"
    compile project(':clitests')
    compile project(':clitests').sourceSets.test.output
}
```

Which refers to the `incubator-openwhisk` tests directory (see `settings.gradle`):

```
include 'clitests'; project(':clitests').projectDir = new File(owDirectory, 'tests')
```

I'm experimenting with tearing out clitests to see what happens...

build.gradle
```diff
@@ -1,93 +1,286 @@
...
+        build(['name':'github.com/mattn/go-isatty', 'version':'66b8e73f3f5cda9f96b69efd03dd3d7fc4a5cdb8', 'transitive':false])
+        build(['name':'github.com/mitchellh/go-homedir', 'version':'1111e456ffea841564ac0fa5f69c26ef44dafec9', 'transitive':false])
+        build(['name':'github.com/nicksnyder/go-i18n/i18n/...', 'version':'37e5c2de3e03e4b82693e3fcb4a6aa2cc4eb07e3', 'transitive':false])
+        build(['name':'github.com/spf13/cobra', 'version':'1238ba19d24b0b9ceee2094e1cb31947d45c3e86', 'transitive':false])
+        build(['name':'github.com/spf13/pflag', 'version':'367864438f1b1a3c7db4da06a2f55b144e6784e0', 'transitive':false])
+        build(['name':'golang.org/x/sys/unix', 'version':'7f918dd405547ecb864d14a8ecbbfe205b5f930f', 'transitive':false])
+        build(['name':'gopkg.in/yaml.v2', 'version':'cd8b52f8269e0feb286dfeef29f8fe4d5b397e0b', 'transitive':false])
+        build(['name':'github.com/ghodss/yaml', 'version':'0ca9ea5df5451ffdf184b4428c902747c2c11cd7', 'transitive':false])
+        build(['name':'github.com/apache/incubator-openwhisk-client-go/whisk','version':'a67e8509a92beb6c68f0c9da43562af1f5d2b13c','transitive':false])
```
dubeejw: Should be `ba3bbee442357a239667ef6de378d5b7d33e0ceb`.

.travis.yml
```diff
@@ -8,30 +8,36 @@ matrix:
...
           services: docker
+env:
+    secure: "Br0fD9CAKm8gqyEuwmltNJd4dGJCxPpj6feugHlO+CFFwfE/+kJKkpTlsDuRfrUzUDlWiETNPf0XGSjvPFqZExnLCE9XQh2+XF6u+S3YBWfM+rbbyRVAK6BlTwmt0u3jRJ2JP7spedTGZA+qfIWI+UkwoOexo7NcqtMPLahiZzheaaad8y3J+crHQCrB/kPrhLqKVMEOkIbveFdfV2QLfCOWgqP8e1LGZhPZ2N4QcNo0iB5uI4ZyYszTZDniXFKxz7kBs4tl4ZQDqRHqL02qKPsjbvjrZp83ql+PbC2dpgXi9YpaDuBEqKKX1rTQP5ppcwbobot5U3ItHWzpXbLCdsWxvbde/0enjMmOF1wwl71hPYIf7PkQmNAWXRtL2Z1TguO/dKCeXBLDER4YDQ79GYpikAMqnrRLou4rsyZrNUzg8aHbELzAHppDSpqEJN8ymGgWgmWBT8yPaWVwN5CjPFOxLPrVEObcwcNSGOcOvtmUGCnXKSZminNZUjz5QfWqpA+LYCUIbeOI2X2D+iQWklWONaU2A+PcNLaaegwwj4K+9ZlU08Ed8Ud5ZGhjN9s72OtoFPdSHAyBgf/qxIdnTzmmE+SQ90mDtG3VqjHy28Ix7cIGqaIEC8shFb0kKHqQ7AVkLPm0bsh2fbbYu/5YVpXHfxlKWfOLBV14bHUq3v0="
+    global:
+        - build_file_name=wsk
```
dubeejw: I don't think this variable is used anymore.

PR #185, opened 2018-01-04
Move routemgmt src and install scripts from main openwhisk repo

ServoKvd:


ServoKvd:
Hi  @csantanapr , @mdeuser 
Could you review this pull request? One thing still to be done is to move all the routemgmt files that I merged from incubator-openwhisk to its own folder here like "routemgmt".
I could move over files, while retaining the history from the main openwhisk repo, using " git filter-branch --subdirectory--filter core/routemgmt -- --all". However if I create a new folder in the cli repo to move these files to - the git history is lost again. So for now all new files are in the root directory of incubator-openwhisk-cli.
Could you suggest how to do ' git mv" within the repo, that preserves history?

dubeejw:
@ServoKvd, I see the CLI is needed for the install scripts. Is it needed for anything else in this PR?

ServoKvd:
Hi @dubeejw, No no-one else uses it  -  we just needed to find a better place for after we decided to move it out of main openwhisk repo. And routemgmt install itself is not a mandatory OW component - as far as I know originally meant as a test interface for api gw?

dubeejw:
After conversations with @mdeuser and @csantanapr it sounds like we are going to host the route management source here at least temporarily.

dubeejw:
@csantanapr, please review.

mdeuser:
is there a co-dependent pr in the openwhisk repo?

if i understand correctly, these changes are sprinkle the apimgmt files in separate folders off the repo's root folder.  i suggest we consolidate these files under a common `apimgmt` folder.

is this a good time to change file names?  `installRouteMgmt.sh` -> `installApiMgmt.sh`.. same for unintall..

rabbah:
These actions belong in the API gw. The CLI just has a canonical reference to an "API" endpoint.

PR #191, opened 2018-01-09
Handle namespace API listing on the client.

### LINKED ISSUE #186: update wsk list to make separate calls to controller to list actions, triggers, rules and packages
`wsk list` currently performs a GET on /namespaces/_. I've proposed removing this API (see https://github.com/apache/incubator-openwhisk/issues/3153).

rabbah:
Replaces call to GET /namespace/_ with calls to GET /namespace/_/[actions, triggers, rules, packages]

For https://github.com/apache/incubator-openwhisk/issues/3153
Fixes #186

### DIFF
--- a/commands/namespace.go
+++ b/commands/namespace.go
@@ -61,50 +61,33 @@ var namespaceListCmd = &cobra.Command{
 }
 
 var namespaceGetCmd = &cobra.Command{
-    Use:   "get [NAMESPACE]",
-    Short: wski18n.T("get triggers, actions, and rules in the registry for a namespace"),
+    Use:   "get",
+    Short: wski18n.T("get triggers, actions, and rules in the registry for namespace"),
     SilenceUsage:   true,
     SilenceErrors:  true,
     PreRunE: SetupClientConfig,
     RunE: func(cmd *cobra.Command, args []string) error {
-        var qualifiedName = new(QualifiedName)
         var err error
+        var namespace string = getClientNamespace()
 
-        if whiskErr := CheckArgs(args, 0, 1, "Namespace get",
-                wski18n.T("An optional namespace is the only valid argument.")); whiskErr != nil {
+        if whiskErr := CheckArgs(args,0,0,"Namespace get",
+                wski18n.T("No arguments are required.")); whiskErr != nil {
             return whiskErr
         }
 
-        // Namespace argument is optional; defaults to configured property namespace
-        if len(args) == 1 {
-            if qualifiedName, err = NewQualifiedName(args[0]); err != nil {
-                return NewQualifiedNameError(args[0], err)
-            }
+        actions, _, err := Client.Actions.List("", &whisk.ActionListOptions{ Skip: 0, Limit: 0 })
+        if err != nil { return entityListError(err, namespace,"Actions") }
 
-            if len(qualifiedName.GetEntityName()) > 0 {
-                return entityNameError(qualifiedName.GetEntityName())
-            }
-        }
+        packages, _, err := Client.Packages.List(&whisk.PackageListOptions{ Skip: 0, Limit: 0 })
+        if err != nil { return entityListError(err, namespace,"Packages") }
 
-        namespace, _, err := Client.Namespaces.Get(qualifiedName.GetNamespace())
+        triggers, _, err := Client.Triggers.List(&whisk.TriggerListOptions{ Skip: 0, Limit: 0 })
+        if err != nil { return entityListError(err, namespace,"Triggers") }
 
-        if err != nil {
-            whisk.Debug(whisk.DbgError, "Client.Namespaces.Get(%s) error: %s\n", getClientNamespace(), err)
-            errStr := wski18n.T("Unable to obtain the list of entities for namespace '{{.namespace}}': {{.err}}",
-                    map[string]interface{}{"namespace": getClientNamespace(), "err": err})
-            werr := whisk.MakeWskErrorFromWskError(errors.New(errStr), err, whisk.EXIT_CODE_ERR_NETWORK,
-                whisk.DISPLAY_MSG, whisk.NO_DISPLAY_USAGE)
-            return werr
-        }
-
-        fmt.Fprintf(color.Output, wski18n.T("Entities in namespace: {{.namespace}}\n",
-            map[string]interface{}{"namespace": boldString(getClientNamespace())}))
-        sortByName := Flags.common.nameSort
-        printList(namespace.Contents.Packages, sortByName)
-        printList(namespace.Contents.Actions, sortByName)
-        printList(namespace.Contents.Triggers, sortByName)
-        //No errors, lets attempt to retrieve the status of each rule #312
-        for index, rule := range namespace.Contents.Rules {
+        rules, _, err := Client.Rules.List(&whisk.RuleListOptions{ Skip: 0, Limit: 0 })
+        if err != nil { return entityListError(err, namespace,"Rules") }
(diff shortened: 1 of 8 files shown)


commands/namespace.go
```diff
@@ -86,25 +86,28 @@ var namespaceGetCmd = &cobra.Command{
             }
         }
 
-        namespace, _, err := Client.Namespaces.Get(qualifiedName.GetNamespace())
-
-        if err != nil {
-            whisk.Debug(whisk.DbgError, "Client.Namespaces.Get(%s) error: %s\n", getClientNamespace(), err)
-            errStr := wski18n.T("Unable to obtain the list of entities for namespace '{{.namespace}}': {{.err}}",
-                    map[string]interface{}{"namespace": getClientNamespace(), "err": err})
+        errHandler := func(err error, kind string) error {
```
dubeejw: Good idea! Maybe it's a good idea to put it in shared.go so other methods can utilize it. Plus it will serve as a template for debug/error handling to reduce code duplication in similar circumstances.

commands/namespace.go
```diff
@@ -86,25 +86,28 @@ var namespaceGetCmd = &cobra.Command{
...
+            whisk.Debug(whisk.DbgError, "Client.%s.Get(%s) error: %s\n", getClientNamespace(), kind, err)
```
dubeejw: Should be `Client.%s.List(%s)` instead of `Client.%s.Get(%s)`.

rabbah:
Thanks - two of these were easy to fix, I reverted the error message.

For this test
```scala
it should "not list entities with an invalid namespace" in {
    val namespace = "fakeNamespace"
    val stderr = wsk.namespace.get(Some(s"/${namespace}"), expectedExitCode = FORBIDDEN).stderr

    stderr should include(s"Unable to obtain the list of entities for namespace '${namespace}'")
  }
```
Why does the CLI even accept a namespace positional argument here? This is independent of the PR FWIW. Should this behavior be disallowed?

rabbah:
@dubeejw what change are you requesting? Moving the errHandler function? I can move it to shared.go but it seems useless to do that without then also refactoring all the error handlers to use it. Maybe I misunderstood.

dubeejw:
@rabbah, `Get` to `List` is the only thing that really need to be changed.

dubeejw:
Moving the error handler would be nice. Refactoring the functions can be done later.

dubeejw:
Seems like we can remove the namespace get test and update the command usage (https://github.com/dubeejw/incubator-openwhisk-cli/blob/master/commands/namespace.go#L64).

@mdeuser, any input on the above?

mdeuser:
the original python based cli implementation took a positional argument, so that syntax was carried forward into the go implementation.
```
C:\>wsk namespace get -h
usage: wsk-script.py namespace get [-h] [-u AUTH] [name]

positional arguments:
  name                  the namespace to list
```

rabbah:
Is this actually implemented in the go version? I suspected it wasn’t but I didn’t dig too deeply.

commands/namespace.go
```diff
@@ -86,25 +86,28 @@ var namespaceGetCmd = &cobra.Command{
             }
         }
 
-        namespace, _, err := Client.Namespaces.Get(qualifiedName.GetNamespace())
-
-        if err != nil {
-            whisk.Debug(whisk.DbgError, "Client.Namespaces.Get(%s) error: %s\n", getClientNamespace(), err)
+        errHandler := func(err error, kind string) error {
+            whisk.Debug(whisk.DbgError, "Client.%s.List(%s) error: %s\n", getClientNamespace(), kind, err)
```
mdeuser: the first `%s` in the `Client.%s.List(%s)` should be the entity name (i.e. Actions, Packages, etc) that matches the actual Client function name - not the namespace argument.

rabbah:
I’ll investigate.

commands/namespace.go
```diff
@@ -86,25 +86,28 @@ var namespaceGetCmd = &cobra.Command{
             }
         }
 
-        namespace, _, err := Client.Namespaces.Get(qualifiedName.GetNamespace())
-
-        if err != nil {
-            whisk.Debug(whisk.DbgError, "Client.Namespaces.Get(%s) error: %s\n", getClientNamespace(), err)
+        errHandler := func(err error, kind string) error {
```
mdeuser: for consistency with other existing error handler defintions, i suggest moving this error handler outside of the `RunE` function definition... in this same module though.

mdeuser:
if a user has permissions to access another namespace then `wsk namespace get NAMESPACE` should work with the go implementation.  
```
>wsk namespace get /whisk.system
error: Unable to obtain the list of entities for namespace 'whisk.system': The supplied authentication is not authorized to access 'whisk.system'. (code 22018494)
```

rabbah:
The behavior dates back to when a single key could access multiple spaces. This is no longer allowed so supporting this (wsk list x) doesn’t make sense anymore in the client.

commands/namespace.go
rabbah: Thanks - order fixed.

mdeuser:
@rabbah - as an aside, what are the current cross-namespace shareable entity types these days?

rabbah:
public packages and nothing else.

commands/namespace.go
```diff
@@ -61,50 +61,35 @@ var namespaceListCmd = &cobra.Command{
...
-
-        // Namespace argument is optional; defaults to configured property namespace
-        if len(args) == 1 {
-            if qualifiedName, err = NewQualifiedName(args[0]); err != nil {
-                return NewQualifiedNameError(args[0], err)
-            }
-
-            if len(qualifiedName.GetEntityName()) > 0 {
-                return entityNameError(qualifiedName.GetEntityName())
+        if (!(len(args) == 1 && args[0] == "/_")) {
```
dubeejw: This seems to clash a little with the usage info that states the namespace parameter is optional.

commands/shared.go
```diff
@@ -33,3 +33,12 @@ func entityNameError(entityName string) (error) {
 
     return whisk.MakeWskError(errors.New(errMsg), whisk.EXIT_CODE_ERR_GENERAL, whisk.DISPLAY_MSG, whisk.DISPLAY_USAGE)
 }
+
+func entityListError(err error, namespace string, kind string) error {
```
dubeejw: 👍

commands/namespace.go
dubeejw: Also, there isn't a test for `wsk namespace get /_`.

rabbah: there is, the CLI tests already do this: `wsk namespace get /_` (that's how I found it).
I have to add support for `/_` because there's a circular dependence between this repo and the main repo and can't break it otherwise.

rabbah:
go deps updated.

As for updating wsk basic tests in the main repo, see https://github.com/apache/incubator-openwhisk/pull/3167.

The tests shouldn't be running in the main repo and will be removed by https://github.com/apache/incubator-openwhisk/pull/3103 and https://github.com/apache/incubator-openwhisk/pull/2996.

If you have an better idea to break the circular dependence I'm all ears.

rabbah:
To clarify, the tests in this repo have a dependence on the Wsk CLI client in the main repo: https://github.com/apache/incubator-openwhisk/blob/master/tests/src/test/scala/common/Wsk.scala#L761

That Wsk CLI generates the `/_`.
So there's no way to make the tests here pass by removing the position argument unless I remove the tests or do anyone of more funky things.

PR #194, opened 2018-01-12
Update README regarding the environment variable OPENWHISK_HOME

houshengbo:
This PR adds the instruction about OPENWHISK_HOME needs to be set
for the gradle build.

### DIFF
--- a/README.md
+++ b/README.md
@@ -60,6 +60,7 @@ If it is executed successfully, you can find your binary `wsk` directly under Op
 
 This is the second choice for you to build the binary. Make sure that you have Docker and gradle on your machine:
 [installing Docker](https://docs.docker.com/engine/installation/) and [installing Gradle](https://gradle.org/install) for your local machine.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -60,6 +60,7 @@ If it is executed successfully, you can find your binary `wsk` directly under Op
 
 This is the second choice for you to build the binary. Make sure that you have Docker and gradle on your machine:
 [installing Docker](https://docs.docker.com/engine/installation/) and [installing Gradle](https://gradle.org/install) for your local machine.
+The environment variable OPENWHISK_HOME needs to be correctly set, by pointing to the home directory of openwhisk.
```
dubeejw: Would use `OpenWhisk` instead of `openwhisk` to make it consistent with the rest of the document.

jonpspri: I added code to the settings.gradle in #184 that would default the OPENWHISK_HOME to '../incubator-openwhisk', which will (often) work, as well as deliver a sane error message.  Ymmv.

PR #197, opened 2018-01-18
Add ppc64le to arch to build cli

sudswas:
This patch adds support for ppc64le architecture for cross compile
builds for the wsk CLI, following the same suite of changes done
for s390x.


jonpspri:
Hi @dubeejw @houshengbo,

I do address Power in the work I'm doing over in #184 , but this is a good short-term addition while we continue with review.

-J

Issue #200, opened 2018-01-20
Investigate why Travis is pulling from master incubator-openwhisk-client-go during build


dubeejw:
Travis is failing as it is picking up changes from incubator-openwhisk-client-go and the companion has not been merged in this repo yet. The Travis build should run against the pinned version of client-go from the Godeps instead of master.

Build log: https://api.travis-ci.org/v3/job/330994467/log.txt

Added the error here:
```
commands/api.go:106: cannot range over api.Swagger.Paths[path] (type *whisk.ApiSwaggerPath)
commands/api.go:211: cannot range over retApi.Swagger.Paths[path] (type *whisk.ApiSwaggerPath)
commands/api.go:580: cannot range over resultApi.Swagger.Paths[path] (type *whisk.ApiSwaggerPath)
commands/api.go:618: cannot range over resultApi.Swagger.Paths[path] (type *whisk.ApiSwaggerPath)
commands/api.go:681: cannot range over resultApi.Swagger.Paths[path] (type *whisk.ApiSwaggerPath)
commands/api.go:715: cannot range over resultApi.Swagger.Paths[path] (type *whisk.ApiSwaggerPath)
```

Fail to build with the 'go build' command.

dubeejw:
build_tag_releases.sh may not be running `go get`.

houshengbo:
@dubeejw  I suspect the failure is not due to the wrong version of client go. The cli should be able to pin to the specified version of client go(not master), since almost day 1 these repo are separate. 

I can work on this issue.

houshengbo:
@dubeejw The build failed because cli has new code, but does not pin to the latest version of client go.
The last commit of client go made the change to swagger: a086445: https://github.com/apache/incubator-openwhisk-client-go/commit/a0864455f7c18db70d93d4dd2bc0b43d2334ed90

houshengbo:
The reason for that the PR passed before, but it failed on master: https://api.travis-ci.org/v3/job/330994467/log.txt, is probable a rebase issue.
when it is still pr, cli has not changed yet. when it is merged, there are several commits merged before this one. @dubeejw

PR #201, opened 2018-01-20
Run Godep tool for build

dubeejw:
Closes https://github.com/apache/incubator-openwhisk-cli/issues/200

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -24,6 +24,7 @@ install:
   - go get -u github.com/golang/lint/golint
   - go get -u github.com/stretchr/testify
   - go get -u github.com/spf13/viper
+  - go get github.com/tools/godep
 
 script:
   - ./tools/travis/test_openwhisk.sh
@@ -48,6 +49,7 @@ before_deploy:
   - go get github.com/mattn/go-isatty
   - export build_file_name=wsk
   - export zip_file_name=OpenWhisk_CLI
+  - ./bin/godep restore -v
   - chmod +x tools/travis/build_tag_releases.sh
   - ./tools/travis/build_tag_releases.sh $build_file_name $zip_file_name
   - export RELEASE_PKG_FILE=$(ls $zip_file_name-*.*)
--- a/Makefile
+++ b/Makefile
@@ -9,13 +9,15 @@ BUILD=`git rev-parse HEAD`
 
 deps:
 	@echo "Installing dependencies"
-	go get -d -t ./...
+	./bin/godep restore -v
+	# go get -d -t ./...
 
 LDFLAGS=-ldflags "-X main.CLI_BUILD_TIME=`date -u '+%Y-%m-%dT%H:%M:%S%:z'`"
 
-updatedeps:
-	@echo "Updating all dependencies"
-	@go get -d -u -f -fix -t ./...
+# Never want to do this ever
+# updatedeps:
+# @echo "Updating all dependencies"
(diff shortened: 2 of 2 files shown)


dubeejw:
@houshengbo, please look at this. It seems that the godep tool is not running for the releases script.

.travis.yml
```diff
@@ -48,6 +49,7 @@ before_deploy:
   - go get github.com/mattn/go-isatty
   - export build_file_name=wsk
   - export zip_file_name=OpenWhisk_CLI
+  - ./bin/godep restore -v
```
dubeejw: For some reason two builds are happening. Why not just build once with travis.yml?????

dubeejw:
https://github.com/apache/incubator-openwhisk-cli/pull/184 will supersede this. Just want a quick fix for now.

.travis.yml
dubeejw: We want to distribute binaries that are actually tested. Rebuild the binaries with the wrong dependencies (not using Godeps) and not testing those binaries is not good.

houshengbo:
build: deps is used to build the binaries. It depends on deps, and deps runs "go get -d -t ./..." to install all the dependencies based on what is in the json file for deps. If the dependency is ready installed, there will be no change, no matter it is older or newer version. If the dependency is not available, it will install the version as json indicates.

One advantage I see in the Makefile is building up the dependencies for each task.

.travis.yml
houshengbo: No need to add this line here. As you see in test_openwhisk.sh, we run make build, so the dependecy deps runs as well. All the godep dependencies are installed by  "go get -d -t ./..." .

dubeejw:
@houshengbo, we want to use the godep tool to install the CLI dependancies instead of `go get`. `godep` allows us to lock the dependency versions into place with ones that we know work and have approval to use legally.

Makefile
```diff
@@ -9,13 +9,15 @@ BUILD=`git rev-parse HEAD`
 
 deps:
 	@echo "Installing dependencies"
-	go get -d -t ./...
+	./bin/godep restore -v
```
houshengbo: This should be changed into godep restore -v

Makefile
```diff
@@ -9,13 +9,15 @@ BUILD=`git rev-parse HEAD`
...
+	# go get -d -t ./...
```
houshengbo: Remove this line

Makefile
```diff
@@ -9,13 +9,15 @@ BUILD=`git rev-parse HEAD`
...
 
 LDFLAGS=-ldflags "-X main.CLI_BUILD_TIME=`date -u '+%Y-%m-%dT%H:%M:%S%:z'`"
 
-updatedeps:
-	@echo "Updating all dependencies"
-	@go get -d -u -f -fix -t ./...
+# Never want to do this ever
+# updatedeps:
+# @echo "Updating all dependencies"
+#	@go get -d -u -f -fix -t ./...
```
houshengbo: Remove Line 17 to Line 20

houshengbo:
@dubeejw I finally found out the issue. You are correct. See the comments.

dubeejw:
Merged https://github.com/apache/incubator-openwhisk-cli/pull/202 instead.

.travis.yml
mdeuser: @dubeejw - good catch.  how many builds are now happening?

PR #202, opened 2018-01-20
Updathe godep dependency installation and client go version

houshengbo:


mdeuser:
@houshengbo - what is the use case for the makefile `updatedeps` target?

PR #231, opened 2018-02-25
Update User Agent Test

dubeejw:
The current test uses `wsk list` to examine the user agent value. This is problematic as `wsk list` calls `wsk rule list`, and `wsk rule list` attempts to get the status of every rule that is listed individually. If a rule that was listed is deleted by the time the CLI gets that rule's status, the test will fail.

Perhaps we need to suppress the error from `wsk rule list`.

Related to https://github.com/apache/incubator-openwhisk-cli/issues/232.


csantanapr:
Travis should be enough

PR #244, opened 2018-03-15
Add --web-secure option to action create/update

mdeuser:
- convenience wrapper around the `require-whisk-auth` annotation
- similar to the `web` option, when used to update an existing action all existing annotations will be preserved IF no other annotation is specified
- syntax is `--web-secure true|false|<string>`
  - when set to true, a random numeric secret is generated for the `require-whisk-auth` annotation
  - when set to false, the `require-whisk-auth` annotation is removed
  - when set to any other value, that value is used as the `require-whisk-auth` annotation value
- `--web-secure` is only valid for web actions
- related to https://github.com/apache/incubator-openwhisk/pull/3388
- depends on https://github.com/apache/incubator-openwhisk-client-go/pull/69

### DIFF
--- a/commands/action.go
+++ b/commands/action.go
@@ -463,109 +463,106 @@ func parseAction(cmd *cobra.Command, args []string, update bool) (*whisk.Action,
 
 func augmentAction(cmd *cobra.Command, args []string, action *whisk.Action, update bool) (*whisk.Action, error) {
 	var err error
-	var actionExisting *whisk.Action = nil
-	var augmentedAction *whisk.Action = action
+	var existingAction *whisk.Action = nil
+	var augmentedAction *whisk.Action = new(whisk.Action)
+	*augmentedAction = *action
 
 	if update {
-		if actionExisting, _, err = Client.Actions.Get(action.Name, DO_NOT_FETCH_CODE); err != nil {
+		if existingAction, _, err = Client.Actions.Get(action.Name, DO_NOT_FETCH_CODE); err != nil {
 			return nil, actionGetError(action.Name, DO_NOT_FETCH_CODE, err)
 		}
 	}
 
 	// Augment the action's annotations with the --web related annotations
 	// FIXME MWD - avoid retrieving existing action TWICE!  once above and once in the webAction() below
-	if augmentedAction, err = augmentWebArg(cmd, args, augmentedAction, actionExisting); err != nil {
+	if augmentedAction, err = augmentWebArg(cmd, args, augmentedAction, existingAction); err != nil {
 		return nil, err
 	}
 
 	// Augment the action's annotations with the --web-secure related annotations
-	if augmentedAction, err = augmentWebSecureArg(cmd, args, augmentedAction, actionExisting); err != nil {
+	if augmentedAction, err = augmentWebSecureArg(cmd, args, action, augmentedAction, existingAction); err != nil {
 		return nil, err
 	}
 
-	whisk.Debug(whisk.DbgInfo, "Augmented action struct: %#v\n", action)
+	whisk.Debug(whisk.DbgInfo, "Augmented action struct: %#v\n", augmentedAction)
 	return augmentedAction, err
 }
 
-func augmentWebArg(cmd *cobra.Command, args []string, action *whisk.Action, actionExisting *whisk.Action) (*whisk.Action, error) {
-	// Retain existing annotations iff the --web option is used
-	preserveAnnotations := action.Annotations == nil
-	var augmentedAction *whisk.Action = action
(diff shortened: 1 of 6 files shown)


mdeuser:
@dubee - for your review.  thanks!

mdeuser:
@dubee - agree.. a doc update will be needed once we all agree on the updated cli syntax

mdeuser:
@dubee - here's a pass at a doc update that should reflect the current cli syntax
https://github.com/mdeuser/openwhisk/commit/b18a5dc9534c0e2ac7c44b0d4b4f841ad85ced1e

commands/action.go
```diff
@@ -442,14 +457,109 @@ func parseAction(cmd *cobra.Command, args []string, update bool) (*whisk.Action,
...
+	var err error
+	preserveAnnotations := action.Annotations == nil
+	var augmentedAction *whisk.Action = new(whisk.Action)
+	*augmentedAction = *action
+
 	if cmd.LocalFlags().Changed(WEB_FLAG) {
-		preserveAnnotations := action.Annotations == nil
-		action.Annotations, err = webAction(Flags.action.web, action.Annotations, qualifiedName.GetEntityName(), preserveAnnotations)
+		if existingAction == nil {
+			augmentedAction.Annotations, err = webAction(Flags.action.web, action.Annotations, action.Name, preserveAnnotations)
```
dubee: If `preserveAnnotations` is true, pass in the action we already retrieved to `webAction()` to avoid fetching it again in there.

commands/action.go
```diff
@@ -442,14 +457,109 @@ func parseAction(cmd *cobra.Command, args []string, update bool) (*whisk.Action,
...
+}
 
-	return action, err
+/*
+ * Return a whisk.Action augmented with --web-secure annotation updates
+ * originalAction:  a action constructed from command line argument
+ * action:          an action constructed from command line args + possible other augmentation (i.e. --web annotations)
+ * existingAction:  on an action update, this is the existing action
+ */
+func augmentWebSecureArg(cmd *cobra.Command, args []string, originalAction *whisk.Action, action *whisk.Action, existingAction *whisk.Action) (*whisk.Action, error) {
```
dubee: Seems like adding even more flags that are sugar for annotations would not be very easy to do with this method.

dubee: For instance, there should be a sugar flag for `-a web-custom-options true|false`, but there would be a lot of boilerplate code just to set that up by the looks of it.

tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
```diff
@@ -48,8 +48,8 @@ import JsonArgsForTests._
 import whisk.http.Messages
 
 /**
- * Tests for basic CLI usage. Some of these tests require a deployed backend.
- */
+  * Tests for basic CLI usage. Some of these tests require a deployed backend.
```
dubee: Are these mostly formatting changes? Hard to tell.

commands/action.go
```diff
@@ -442,14 +457,109 @@ func parseAction(cmd *cobra.Command, args []string, update bool) (*whisk.Action,
...
+
+	whisk.Debug(whisk.DbgInfo, "Augmented action struct: %#v\n", augmentedAction)
+	return augmentedAction, err
+}
+
+func augmentWebArg(cmd *cobra.Command, args []string, action *whisk.Action, existingAction *whisk.Action) (*whisk.Action, error) {
+	var err error
+	preserveAnnotations := action.Annotations == nil
+	var augmentedAction *whisk.Action = new(whisk.Action)
+	*augmentedAction = *action
```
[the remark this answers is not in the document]
mdeuser: @dubee good question...   not really.  in general, there are three separate "action" states that must be kept separate

1. the original cli input that generated an action.  this action has the nicely parsed annotations/parameters
2. the in-progress action which may or may not have been manipulated by other command line option processing
3. the existing action (when present)

the option augmentation logic is free to update the in-progress state, but not touch the original or existing action states.  this way each of the special command option processing will have all of the state to make its action augmentation decisions.

commands/action.go
mdeuser: right.  since that logic for any new sugar flag is unique, each new flag would need to implement it's own "augmentation" function.  perhaps in another pr, there might be a table of these augmentation functions which are called in order during action create/update.

mdeuser: not sure.. i think each flag might be so unique that the operations don't boil down to the same augmentation "type".  that said, if at some point we go with the augmentation table approach, if a new sugar flag arrives, it might be able to re-use one of the existing augmentation functions..  the function signature would need enough generalization to allow this though..

PR #246, opened 2018-03-20
Fix Go regex for path parameter parsing

### LINKED ISSUE #237: API with back-to-back path parameters fails
Creating an API with multiple back-to-back path parameters fails
```
>bx wsk api create /dumpurl /two/{segment1}/{segment2} get showPathSegs --response-type http
error: Unable to create API: API creation failure: The parameter 'segment2' defined in path '/two/{segment1}/{segment2}' does not match any of the parameters defined for the path in the swagger file.
```
Multiple paths separated by a non-parameter path segment work correctly
```
>bx wsk api create /dumpurl /two/{segment1}/and/{segment2} get showPathSegs --response-type http
ok: created API /dumpurl/two/{segment1}/and/{segment2} GET for action /_/showPathSegs
https://service.us.apiconnect.ibmcloud.com/gws/apigateway/api/XXXXXXXXXX/dumpurl/two/{segment1}/and/{segment2}
```
Fix is needed for the following
1. Support back-to-back path segments
2. Error message in the context of the failing command does not make sense... that is, no swagger file was specified in the command, yet the error message indicates that a swagger file is in error.

mdeuser:
Update Go regex code used to extract path parameters
Update test to have a back-to-back path parameter

Fixes #237

### DIFF
--- a/commands/api.go
+++ b/commands/api.go
@@ -97,20 +97,34 @@ func isValidRelpath(relpath string) (error, bool) {
 	return nil, true
 }
 
-func hasPathParameters(path string) (bool, error) {
-	matched, err := regexp.MatchString(pathParamRegex, path)
-	hasBracket := strings.ContainsRune(path, '{') || strings.ContainsRune(path, '}')
+func getPathParameterNames(path string) ([]string, error) {
+	var pathParameters []string
+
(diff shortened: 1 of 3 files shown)


mdeuser:
@drcariel -- would you mind reviewing

PR #247, opened 2018-03-20
Fix intermittent test failure

dubee:
`create, and get a trigger summary` fails intermittently as the returned annotations array is not always in the expected order.

### DIFF
--- a/tests/src/test/scala/system/basic/WskBasicTests.scala
+++ b/tests/src/test/scala/system/basic/WskBasicTests.scala
@@ -549,13 +549,18 @@ class WskBasicTests extends TestHelpers with WskTestHelpers {
 
     result.fields("name") shouldBe name.toJson
     result.fields("namespace") shouldBe ns.toJson
-    result.fields("annotations") shouldBe JsArray(
+    val receivedAnnotations = result.fields("annotations").convertTo[JsArray].elements
+    val expectedAnnotations = JsArray(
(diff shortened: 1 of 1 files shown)


tests/src/test/scala/system/basic/WskBasicTests.scala
```diff
@@ -549,13 +549,18 @@ class WskBasicTests extends TestHelpers with WskTestHelpers {
...
+    val expectedAnnotations = JsArray(
       JsObject("key" -> JsString("description"), "value" -> JsString("Trigger description")),
       JsObject(
         "key" -> JsString("parameters"),
         "value" -> JsArray(
           JsObject("description" -> JsString("Parameter description 1"), "name" -> JsString("paramName1")),
-          JsObject("description" -> JsString("Parameter description 2"), "name" -> JsString("paramName2")))))
+          JsObject("description" -> JsString("Parameter description 2"), "name" -> JsString("paramName2"))))).elements
+
+    for (expectedItem <- expectedAnnotations) {
```
mdeuser: would this also work?  maybe a bit clearer
```expectedAnnotations should contain theSameElementsAs receivedAnnotations ```

dubee: @mdeuser, giving it a try.

dubee:
@mdeuser, anything else on this one?

PR #251, opened 2018-03-20
Add the support to specify the version of the binaries

houshengbo:
This PR changes the task name from release to releaseBinaries in
order to make it consistent in the release process. It also adds
packageVersion as a parameter to specify the version of the binaries
to be released.

Partially-closes: apache/incubator-openwhisk-release#9

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -61,7 +61,6 @@ class OpenWhiskPlatform {
         The 'zipFileName' property is the root file name to use for archives.
      */
     static String zipFileName
-    static String packageVersion
 
     /*
         Create a platform for the local platform
@@ -122,9 +121,8 @@ OpenWhiskPlatform.zipFileName =
     System.env['zip_file_name'] ?:
         (rootProject.findProperty('zipFileName') ?: 'OpenWhisk_CLI')
 
-OpenWhiskPlatform.packageVersion =
-        System.env['packageVersion'] ?:
-                (rootProject.findProperty('packageVersion') ?: 'latest')
+project.ext.packageVersion =
+        rootProject.findProperty('packageVersion') ?: 'latest'
 
 String buildFileName = System.env['build_file_name'] ?:
         (rootProject.findProperty('buildFileName') ?: 'wsk')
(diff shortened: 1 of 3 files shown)


houshengbo:
@jonpspri Please help me review this PR, thanks.

build.gradle
```diff
@@ -61,6 +61,7 @@ class OpenWhiskPlatform {
         The 'zipFileName' property is the root file name to use for archives.
      */
     static String zipFileName
+    static String packageVersion
```
jonpspri: I think this belongs a project extension rather than an attribute of OpenWhiskPlatform.  Platform really is intended to deal solely with the process/os aspects of the build and is less volatile than version.

Additional thought on why this is important — I think it would be a Bad Idea (but possible and implicitly supported) to build different version numbers for different architectures if we set the `packageVersion` for each platform.

That means taking off the `OpenWhisk.` and `p.` below.

build.gradle
```diff
@@ -121,6 +122,10 @@ OpenWhiskPlatform.zipFileName =
     System.env['zip_file_name'] ?:
         (rootProject.findProperty('zipFileName') ?: 'OpenWhisk_CLI')
 
+OpenWhiskPlatform.packageVersion =
+        System.env['packageVersion'] ?:
+                (rootProject.findProperty('packageVersion') ?: 'latest')
```
jonpspri: Per my comment above, this can become `project.ext.packageVersion =`.  Also, I’m not sure I’d use an environment variable because it hides the “I’m building a release now” from the command line.  If we have a variable set in a Travis build, we should pass it on the command line as the build parameter (as you provide here).

build.gradle
houshengbo: Thx. Done.

build.gradle
houshengbo: Done. I remove the configuration as a system env.

houshengbo:
@jonpspri Thank you. I have changed this PR based on the comments.

jonpspri:
LGTM

PR #256, opened 2018-03-23
add webhook for slack notification of PR test results

dgrove-oss:


dgrove-oss:
hook verified.
screen shot 2018-03-27 at 5 39 08 pm

Issue #259, opened 2018-03-26
For Ease of Maintenance, Update and Decide Where these Tests Should Live


drcariel:
There is currently a level of manual syncing going on between duplicate test files that exist in both `https://github.com/apache/incubator-openwhisk-cli` and `https://github.com/apache/incubator-openwhisk`. I noticed the following files, sharing the same name unless indicated otherwise, that could use further investigation and/or updating:

- [x] `ApiGwEndToEndTests.scala`
- [x] `WskActionTests.scala`
- [x] `WskBasicJavaTests.scala`
- [x] `WskBasicNodeTests.scala` (Moved CLI versions from open to cli)
- [x] `WskBasicPythonTests.scala`
- [x] `WskBasicSwift3Tests.scala`
- [x] `WskCliBasicNodeTests.scala` (`WskCliBasicNodeDefaultTests.scala` and extended in other files to use both node 6  and 8)
- [x] `WskConsoleTests.scala`
- [x] `WskPackageTests.scala`
- [x] `WskRuleTests.scala` (diverged by 10 lines)
- [x] `WskSequenceTests.scala`
- [x] `ApiGwRoutemgmtActionTests.scala`
- [x] `BaseApiGwTests.scala`
- [x] `JsonArgsForTests.scala` (changed to depend on `TestJsonArgs.scala` in open)
- [x] `WskActionSequenceTests.scala`
- [x] `WskEntitlementTests.scala`
- [x] `WskWebActionsTests.scala` (diverged by 6 lines)
- [ ] ~`ApiGwTests.scala`~ (renamed to indicate rest vs cli in respective repos)
- [ ] ~`WskBasicUsageTests.scala`~ (renamed to indicate rest vs cli in respective repos)
- [ ] ~`WskBasicTests.scala`~ (renamed to indicate rest vs cli in respective repos)

Updating the test builds to rely on the original copies of most of the above that exist in `incubator-openwhisk` will be much easier to maintain. Which tests belong in this repo and vice versa?

drcariel:
It seems to me that they are intended to be identical, however some seem to have diverged, as mentioned in my list above. There may be more divergences as I have only glanced at all the tests that seem identical at first.

drcariel:
I will be working to remove the baseWsk versions from open-cli since we have the sources provided via openwhisk maven jar deps

drcariel:
[...]

PR #262, opened 2018-03-29
rely on single base test versions from incubator openwhisk

drcariel:
sister PR: https://github.com/apache/incubator-openwhisk/pull/3503

issue https://github.com/apache/incubator-openwhisk-cli/issues/259

see issue for more details


drcariel:
The one test that may still fail is `system.basic.WskCliSequenceTests ` I will need to figure out what to do here.

Anyone know much about why WhiskConfig would fail when using the Maven jars? or even after overriding?

tests/src/test/scala/system/basic/WskCliActionTests.scala
```diff
@@ -20,9 +20,33 @@ package system.basic
...
+  }
+
+  it should "not be able to use --kind and --docker at the same time when running action create" in {
+    val file = TestCLIUtils.getTestActionFilename(s"echo.js")
+    val out = wsk.action.create(name = "kindAndDockerAction", artifact = Some(file), expectedExitCode = NOT_ALLOWED,
+        kind = Some("nodejs:6"), docker = Some("mydockerimagename"))
+    out.stderr should include("Cannot specify both --kind and --docker at the same time")
+  }
+
+  it should "not be able to use --kind and --docker at the same time when running action update" in {
```
dubee: Looks like this test can be folded into the above test allowing variable reuse.

dubee: Generally I've been using `Seq` to test permutations like below:
```
Seq(false, true).foreach { updateValue =>
  val out = wsk.action.create(name = "kindAndDockerAction", artifact = Some(file), expectedExitCode = NOT_ALLOWED, kind = Some("nodejs:6"), docker = Some("mydockerimagename"), update = updateValue)
  out.stderr should include("Cannot specify both --kind and --docker at the same time")
}
```

tests/src/test/scala/system/basic/WskCliRuleTests.scala
```diff
@@ -22,7 +22,35 @@ import org.scalatest.junit.JUnitRunner
...
 import common.Wsk
 
+import spray.json._
+import spray.json.DefaultJsonProtocol._
+
 @RunWith(classOf[JUnitRunner])
 class WskCliRuleTests extends WskRuleTests {
   override val wsk: Wsk = new Wsk
+
+  it should "preserve rule status when a rule is updated" in withAssetCleaner(wskprops) { (wp, assetHelper) =>
```
dubee: Does this test need to different from the one in https://github.com/apache/incubator-openwhisk/pull/3503/files?

drcariel: see https://github.com/apache/incubator-openwhisk/pull/3503/files#r178551945

I can merge the tests w/ some logic if we find that is more appropriate

tests/src/test/scala/system/basic/WskCliActionTests.scala
drcariel: done

tests/src/test/scala/system/basic/WskCliRuleTests.scala
drcariel: I have merged them, see https://github.com/apache/incubator-openwhisk/pull/3503/files#diff-907d34bcf62c94014993d5450a69ba23R106

tests/src/test/scala/system/basic/WskCliActionTests.scala
```diff
@@ -20,9 +20,21 @@ package system.basic
...
+import common.TestCLIUtils
+import common.TestUtils.NOT_ALLOWED
 import common.Wsk
 
 @RunWith(classOf[JUnitRunner])
 class WskCliActionTests extends WskActionTests {
   override val wsk = new Wsk
+  override val cli = true
+
+  it should "not be able to use --kind and --docker at the same time when running action create or update" in {
```
dubee: Why add this test here instead of in the base class?

tests/src/test/scala/system/basic/WskCliBasicNodeTests.scala
```diff
@@ -23,6 +23,6 @@ import org.junit.runner.RunWith
 import org.scalatest.junit.JUnitRunner
 
 @RunWith(classOf[JUnitRunner])
-class WskCliBasicNodeTests extends WskBasicNodeTests {
+class WskCliBasicNodeDefaultTests extends WskBasicNodeDefaultTests {
```
dubee: Why the name change to default?

tests/src/test/scala/whisk/core/cli/test/WskBasicUsageTests.scala
```diff
@@ -45,7 +45,7 @@ import whisk.core.entity.MemoryLimit._
 import whisk.core.entity.TimeLimit._
 import whisk.core.entity.size.SizeInt
 import whisk.utils.retry
-import JsonArgsForTests._
+import whisk.core.cli.test.TestJsonArgs._
```
dubee: 👍

rabbah:
> I am wondering if we aren't going to test APIGW functionality in the CLI anymore as those are the only tests I do not see extended.

Isn’t that the only place really to test the api gw? Since from the backend point of view they’re just actions.

mdeuser:
need to retain api gw testing with the cli for now.

drcariel:
lastly, `class ApiGwCliEndToEndTests extends ApiGwEndToEndTests` so we are not loosing any coverage here @dubee

mdeuser:
@drcariel - since api gw functionality is primarily with the openwhisk-cli with the exception of the related web actions (which should eventually move as well i suppose), the `ApiGwEndToEndTests` should move to the openwhisk-cli repo.

drcariel:
@mdeuser are you sure it is appropriate to host the `baseWsk` version of `ApiGwEndToEndTests.scala` here? I think it is better left in `incubator-openwhisk` since `ApiGwRestEndToEndTests.scala` extends `ApiGwEndToEndTests` in that repo

tests/src/test/scala/system/basic/WskCliSequenceTests.scala
```diff
@@ -29,4 +31,5 @@ import common.Wsk
 @RunWith(classOf[JUnitRunner])
 class WskCliSequenceTests extends WskSequenceTests {
   override val wsk: common.Wsk = new Wsk
+  override val whiskConfig = new WhiskConfig(Map(WhiskProperties.getProperty("limits.actions.sequence.maxLength") -> null))
```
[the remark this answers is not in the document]
mdeuser: without hardcoded file path..?

tests/src/test/scala/system/basic/WskCliRuleTests.scala
```diff
@@ -25,4 +25,5 @@ import common.Wsk
 @RunWith(classOf[JUnitRunner])
 class WskCliRuleTests extends WskRuleTests {
   override val wsk: Wsk = new Wsk
+  //override val cli = true
```
dubee: Need this one?

PR #267, opened 2018-04-05
--web-secure generated integer must be within api gw max int range

mdeuser:


commands/action.go
```diff
@@ -1194,10 +1195,12 @@ func printSavedActionCodeSuccess(name string) {
...
 }
 
-// Generate a random int64 number to be used as a web action's
+// Generate a random number to be used as a web action's require-whisk-auth secret
 func genWebActionSecureKey() int64 {
 	r := rand.New(rand.NewSource(time.Now().Unix()))
-	return r.Int63()
+
+	// Truncate integer for API GW interoperability
+	return r.Int63() & MAX_JS_INT
```
dubee: Is it possible to test this?

mdeuser: not well.  i think the better validation test might be in the api creation (would be a separate pr) to validate this value before using it to create the api.

rabbah: Why restrict this to a 64 bit int? Shouldn’t it be more generically a string of arbitrary length (32 or 64 or 128 characters)?

mdeuser: in general, i think the more popular usage for explicitly provided secret values will be string values, but that's just my opinion :-). regarding numeric support, you are right, json does not really specify a bit limit; however, language support for json integers does impose precision limits.  in this case, the api gw web actions are written in javascript with 53 bit integer precision, which i think may be sufficient for most use cases - in particular for the auto-generated random secret number; if there is need to have greater precision i suppose the user specified number can be quoted and treated as a string.
https://stackoverflow.com/questions/13502398/json-integers-limit-on-size

PR #270, opened 2018-04-06
use utils from open and rename duplicate tests to be accurate

drcariel:
- Gets rid of TestCLIUtils entirely to rely on TestUtils in open repo.
- Moves timestamp method to sed TestUtils
- ~moves ApigwTests to open repo (if possible)~
- Renames Basic _and apigwtests_ tests to reflect what is being tested (the Cli)

Companion to https://github.com/apache/incubator-openwhisk/pull/3531


drcariel:
tests will not compile until companion is merged

drcariel:
ugh, so I should go add timeout.js back to `incubator-openwhisk`? or just not make any changes to `CLIUtils`?

I feel like ideally we should use a single test Utils helper and a single copy of each action code

drcariel:
I've created https://github.com/apache/incubator-openwhisk/pull/3545 to add it back. I am also removing the `tests\dat\actions` and `test\dat\apigw` test dirs

rabbah:
Wait - why are all the samples being deleted here?

rabbah:
Oh never mind this is the cli repo. Good move.

drcariel:
The only files that do not exist in `incubator-openwhisk` are `testswaggerdoc2V2` and `testswaggerdoc1V2`. What are they used for?

PR #271, opened 2018-04-13
Add the licenses of the dependencies into LICENSE

houshengbo:
Partially-closes: apache/incubator-openwhisk-release#109


daisy-ycguo:
@houshengbo I think we need to copy the LICENSE files of the dependencies to our source code distribution either by including the whole LICENSE content in our LICENSE.txt file, or add a file and refer to that file in LICENSE.txt. Please read the policy at Apache [here](http://www.apache.org/legal/release-policy.html#license-file).

houshengbo:
@daisy-ycguo I have added a folder named licenses to host all the license content, with the references in the LICENSE.txt file.

houshengbo:
@daisy-ycguo After I added the licenses of dependencies, I think the current contents of NOTICE and LICENSE have something overlapping. For example, they both explain the licenses of the dependencies.

To me, other packages, like Pflag, unix, sys, etc, are the source code dependencies, not the existing components of the project, so we need to remove them from the notice, and keep them in the LICENSE. 

One example for the component of the project can be openwhisk core needs to deploy kafka, zookeeper services, which are not part of the scala source code, but the necessity of the services. They are components of the openwhisk core service.

What do you think?

PR #273, opened 2018-04-16
Fix spray-json import to comply with newer versions.

markusthoemmes:
Needed to upgrade spray versions to 1.3.4. Refer to https://github.com/spray/spray-json/releases/tag/v1.3.4.


tests/src/test/scala/system/basic/WskCliBasicTests.scala
```diff
@@ -34,7 +34,7 @@ import common.WskProps
 import common.WskTestHelpers
 import spray.json._
 import spray.json.DefaultJsonProtocol._
-import spray.json.pimpAny
+import spray.json._
```
dubee: Checkout line 35 `import spray.json._` is already defined

PR #276, opened 2018-04-17
allow exclusion of api_host validation

drcariel:
certain environments may return router IP address instead of api_host string causing a failure
Overriding noApiHostCheck to true will avoid the apihost check


tests/src/test/scala/whisk/core/cli/test/WskCliBasicUsageTests.scala
```diff
@@ -555,7 +558,9 @@ class WskCliBasicUsageTests extends TestHelpers with WskTestHelpers {
     withActivation(wsk.activation, run) { activation =>
       activation.response.status shouldBe "success"
       val fields = activation.response.result.get.convertTo[Map[String, String]]
-      fields("api_host") shouldBe WhiskProperties.getApiHostForAction
+      if (!noApiHostCheck) {
```
mdeuser: @drcariel - nit.. double negative is a littler harder to read that `if (apiHostCheck) {`

PR #277, opened 2018-04-17
Remove JSON.parseFull references

dubee:
Remove `JSON.parseFull` references as it is deprecated causing Travis to fail.


dubee:
@mdeuser, can you review this one? This should fix the Travis breakage.

mdeuser:
@dubee - lgtm.  only question i have is in regards to the `catch` removal.  in theory, it might have caught some unknown exception during the test and gracefully failed the test.. not sure why it's here and not in other tests though.

Issue #291, opened 2018-05-03
Activation log --strip command malforms output


dubee:
When displaying activation logs with the `--strip` flag, the first 39 characters of a log line are removed to exclude the log timestamp and pipe. Removing the first 39 characters is not ideal as log timestamp length can differ resulting in malformed output.

Example of `wsk activation logs` with timestamp length less than 39 characters.
```
$ wsk activation logs 40a2571069c94bc8a2571069c9bbc80f
2018-05-02T19:33:32.829992819Z stdout: this is stdout
2018-05-02T19:33:32.83017613Z stderr: this is stderr
```

Example of malformed output using the `--strip` flag.
```
$ wsk activation logs 40a2571069c94bc8a2571069c9bbc80f --strip
this is stdout
his is stderr
```

Line to be updated:
https://github.com/apache/incubator-openwhisk-cli/blob/1c241dcb2b9ea2afe9dcece690d8e305634e034e/commands/util.go#L286

PR #292, opened 2018-05-03
Account for variable length timestamps when stripping logs

dubee:
See issue for details.

Closes https://github.com/apache/incubator-openwhisk-cli/issues/291

### DIFF
--- a/commands/util.go
+++ b/commands/util.go
@@ -282,7 +282,7 @@ func printFullActivationList(activations []whisk.Activation) {
 
 func printStrippedActivationLogs(logs []string) {
 	for _, log := range logs {
-		regex := regexp.MustCompile("[a-zA-Z0-9\\s]+(stdout|stderr):\\s(.*)")
+		regex := regexp.MustCompile("[a-zA-Z0-9\\s]*(stdout|stderr):\\s(.*)")
 		match := regex.FindStringSubmatch(log)
 
 		if len(match) > 2 && len(match[2]) > 0 {
--- a/commands/util.go
+++ b/commands/util.go
@@ -283,7 +283,7 @@ func printFullActivationList(activations []whisk.Activation) {
 func printActivationLogs(logs []string) {
 	for _, log := range logs {
 		if Flags.activation.strip {
-			fmt.Printf("%s\n", log[39:])
+			fmt.Printf("%s\n", log[strings.Index(log, ": ")+2:])
 		} else {
 			fmt.Printf("%s\n", log)
 		}
--- a/commands/util.go
+++ b/commands/util.go
@@ -280,21 +280,22 @@ func printFullActivationList(activations []whisk.Activation) {
 	}
 }
 
-func printActivationLogs(logs []string) {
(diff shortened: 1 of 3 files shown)


commands/util.go
```diff
@@ -283,7 +283,7 @@ func printFullActivationList(activations []whisk.Activation) {
 func printActivationLogs(logs []string) {
 	for _, log := range logs {
 		if Flags.activation.strip {
-			fmt.Printf("%s\n", log[39:])
+			fmt.Printf("%s\n", log[strings.Index(log, ": ")+2:])
```
mdeuser: or being completely timestamp format agnostic.. ready to ignore future timestamps with embedded ": "
```
log[strings.Index(log, "stdout: ")+strings.Index(log, "stderr: ")+9:]
```

dubee: @mdeuser, would your example break if the user logs `stderr: ` and the pipe being used is `stdout: `, or vice versa?

Ex:
```
"2018-05-02T19:33:32.829992819Z stdout: this is stdout stderr: "
```

mdeuser: good catch.. yeah.. my code example was too compact
```
skip_index := strings.Index(log, "stdout: ")
if skip_index == -1 {
  skip_index = strings.Index(log, "stdout: ")
}
fmt.Printf("%s\n", log[skip_index+8:])
```

dubee: @mdeuser, I think that would result in the same type of problem.

https://play.golang.org/p/cTGf8yO4Krw

mdeuser: https://play.golang.org/p/8zcL24npnPU

dubee: https://play.golang.org/p/ffJ2Vo41S-6

commands/util.go
```diff
@@ -280,14 +280,22 @@ func printFullActivationList(activations []whisk.Activation) {
...
-func printActivationLogs(logs []string) {
+func printStrippedActivationLogs(logs []string) {
 	for _, log := range logs {
-		if Flags.activation.strip {
-			fmt.Printf("%s\n", log[39:])
+		regex := regexp.MustCompile("[a-zA-Z0-9\\s]*(stdout|stderr):\\s(.*)")
+		match := regex.FindStringSubmatch(log)
+
+		if len(match) > 2 && len(match[2]) > 0 {
+			fmt.Printf("%s\n", match[2])
```
mdeuser: what do you think about using a utility function for the individual log strip logic
```
func stripTimestamp(log string) strippedLog string {
    regex := regexp.MustCompile("[a-zA-Z0-9\\s]*(stdout|stderr):\\s(.*)")
    match := regex.FindStringSubmatch(log)
    if len(match) > 2 && len(match[2]) > 0 {
        strippedLog = match[2]
    } else {
        strippedLog = log
    }
    return strippedLog
}
```
then the print function would become
```
func printStrippedActivationLogs(logs []string) {
	for _, log := range logs {
		fmt.Printf("%s\n", stripTimestamp(log))
 	}
 }
```
the unit test would just validate the utility function.

PR #299, opened 2018-05-17
Wait for 2 minutes for logs.

cbickel:
Increase time to wait for logs, to handle potential delays, that could occure, depending on the store, that is used to store logs.


tests/src/test/scala/whisk/core/cli/test/WskCliBasicUsageTests.scala
```diff
@@ -400,7 +400,7 @@ class WskCliBasicUsageTests extends TestHelpers with WskTestHelpers {
             be("this is stdout\nthis is stderr\n") or
               be("this is stderr\nthis is stdout\n")
           }
-        }, 10, Some(1.second))
```
rabbah: 2 minutes - yikes.

dubee: @rabbah, super yikes.

Issue #301, opened 2018-05-17
Display appropriate message when an activation ID is not returned from a trigger fire.


dubee:
Executing a trigger returns an activation ID only when there is rule linked to the trigger. As a result, the CLI should not display `with id` when an activation ID is not returned from a fired trigger.

```
$ wsk trigger create a
ok: created trigger a
$ wsk trigger fire a
ok: triggered /_/a with id 
```

mdeuser:
and should probably pass along a message that indicates the trigger was not really fired.. 
- no rule associate with trigger
- no action associated with trigger's rule (need to confirm this one)

PR #311, opened 2018-06-04
Display appropriate message for trigger fire when 204 is returned

dubee:
When a trigger without active or associated rule(s) is fired, an activation ID is not returned from an OpenWhisk server. These changes display an appropriate message when such an scenario occurs.

Closes https://github.com/apache/incubator-openwhisk-cli/issues/301


commands/trigger.go
```diff
@@ -92,13 +92,24 @@ var triggerFireCmd = &cobra.Command{
 			return werr
 		}
 
+		if resp.StatusCode == 204 {
```
mdeuser: ```if resp.StatusCode == http.StatusNoContent {```

wski18n/resources/en_US.all.json
```diff
@@ -625,6 +625,10 @@
     "id": "{{.ok}} triggered /{{.namespace}}/{{.name}} with id {{.id}}\n",
     "translation": "{{.ok}} triggered /{{.namespace}}/{{.name}} with id {{.id}}\n"
   },
+  {
+    "id": "trigger /{{.namespace}}/{{.name}} did not fire as it is not associated with an active rule(s)\n",
```
mdeuser: i think `rules(s)` can be changed to `rule`.  only one active rule is needed.

tests/src/test/scala/system/basic/WskCliBasicTests.scala
```diff
@@ -744,6 +744,20 @@ class WskCliBasicTests extends TestHelpers with WskTestHelpers {
...
+  it should "display proper error when trigger is not associated with active rule" in withAssetCleaner(wskprops) {
+    val triggerName = withTimestamp("noRuleTrigger")
+
+    (wp, assetHelper) =>
+      assetHelper.withCleaner(wsk.trigger, triggerName) { (trigger, _) =>
+        trigger.create(triggerName)
+      }
+
+      val rr = wsk.trigger.fire(triggerName)
+      val ns = wsk.namespace.whois()
```
mdeuser: not used?

commands/trigger.go
```diff
@@ -92,13 +92,24 @@ var triggerFireCmd = &cobra.Command{
...
+			fmt.Fprintf(color.Output,
+				wski18n.T("trigger /{{.namespace}}/{{.name}} did not fire as it is not associated with an active rule(s)\n",
+					map[string]interface{}{
+						"namespace": boldString(qualifiedName.GetNamespace()),
+						"name":      boldString(qualifiedName.GetEntityName())}))
+
+			return nil
```
mdeuser: up for discussion... is this situation considered an error (i.e. non-zero exit code, stderr)?

dubee: @mdeuser, I believe we agreed to return a non-zero exit code here.

commands/trigger.go
```diff
@@ -92,13 +93,24 @@ var triggerFireCmd = &cobra.Command{
 			return werr
 		}
 
+		if resp.StatusCode == http.StatusNoContent {
+			fmt.Fprintf(color.Output,
+				wski18n.T("trigger /{{.namespace}}/{{.name}} did not fire as it is not associated with an active rule\n",
```
rabbah: I suggest for all new message and to start transitioning these message ids to constants - see for example https://github.com/apache/incubator-openwhisk-cli/pull/315/commits/baa07f047767151c795a1caeaa6743c8370e8a9f.

commands/trigger.go
```diff
@@ -92,13 +93,24 @@ var triggerFireCmd = &cobra.Command{
 			return werr
 		}
 
+		if resp.StatusCode == http.StatusNoContent {
```
rabbah: I suggest starting to add go unit tests as well for such changes as it will vector toward better factoring of the code. For example a unit test that mocks an HTTP response and confirms returning the expected message and CLI code.

commands/trigger.go
mdeuser: @dubee yeah.. that's what you and i agreed upon.  was looking if anyone had an opposite argument :-)

rabbah: which exit code would you return?

PR #327, opened 2018-06-15
Fix failing config test

dubee:
`fail to show api build when setting apihost to bogus value` is failing as it should expect a `NETWORK_ERROR_EXIT` exit code.

Result of retry logic for non-expected network errors: https://github.com/apache/incubator-openwhisk/commit/f4d8ce5009381cdf3385a69c383eca2fd01a0282.

### DIFF
--- a/tests/src/test/scala/whisk/core/cli/test/WskConfigTests.scala
+++ b/tests/src/test/scala/whisk/core/cli/test/WskConfigTests.scala
@@ -46,7 +46,7 @@ class WskConfigTests extends TestHelpers with WskTestHelpers {
     try {
       val env = Map("WSK_CONFIG_FILE" -> tmpwskprops.getAbsolutePath())
       wsk.cli(Seq("property", "set", "-i", "--apihost", "xxxx.yyyy"), env = env)
-      val rr = wsk.cli(Seq("property", "get", "--apibuild", "-i"), env = env, expectedExitCode = ANY_ERROR_EXIT)
+      val rr = wsk.cli(Seq("property", "get", "--apibuild", "-i"), env = env, expectedExitCode = NETWORK_ERROR_EXIT)
       rr.stdout should include regex ("""whisk API build\s*Unknown""")
       rr.stderr should include regex ("Unable to obtain API build information")
     } finally {


rabbah:
In other words should the fix be in the harness - any error exit shouldn’t fail because of a different exit code.

rabbah:
My point is there is a bug here somewhere. If the test specified that it doesn’t care about the exit code why is it failing in code 3?

dubee:
We've had instances were random network errors occurred on the machine running tests. To get around that problem retries were added when `NETWORK_ERROR_EXIT ` exit code is returned but not as an expected exit code. Tests relying on `ANY_ERROR_EXIT` will not be able to take advance of that retry logic.

rabbah:
This test doesn’t need a retry though it’s trying to access an deliberately bad host. 

If the test specified that it doesn’t care about the exit code the test harness should not be checking the exit code. I’ll have a look at that code to get a better idea of what’s going there. I think before chasing this test we should address the underlying mechanics of the harness.

dubee:
Sure this test doesn't need the retry logic, but others may.

rabbah:
I’m not following. Are you disagreeing: a test which specifies that it doesn’t care about the exit code should not fail on exit code 3?

Perhaps there should be a separate flag to retry in network errors.

houshengbo:
To me, the logic seems to become strict, since it does not accept any error code as defined, but only verify the exact error code.

dubee:
I think a test that does not expect a non-zero exit could should explicitly state which exit code it expects.

dubee:
If we prevent the network errors retries for `ANY_ERROR_EXIT` or `DONTCARE_EXIT`, I will have to go through and change most of the tests that use those anyway so that the retry logic works.

rabbah:
This seems like a conflation of concerns and a change in design. We have many tests which use the don’t care code (because they don’t care, rightfully like all the sanitize steps) and makes all tests depends on the cli. If you want to enforce particular exit codes, then write specific tests for them.

rabbah:
Why can’t we support retries and also ignore the final status code on don’t care?

dubee:
We can do that, will just need to update some tests that should actually care what the exit code is. Most of those tests involve network activity.

dubee:
For sanitizing it seems that a retry might help when a network error occurs.

rabbah:
Precisely. Retry may be fine but the  eventual code (404/1 or 200/0) doesn’t matter.

PR #328, opened 2018-06-15
WIP: Fix the Travis slow building issue

houshengbo:


.travis.yml
```diff
@@ -59,14 +59,15 @@ before_script:
...
-  - ./gradlew install
+#  - cd $TRAVIS_BUILD_DIR/../incubator-openwhisk
+#  - ./gradlew install
   - cd $TRAVIS_BUILD_DIR
   - export BUILD_VERSION="latest"
   - if [ ! -z "$TRAVIS_TAG" ] ; then
         export BUILD_VERSION=$TRAVIS_TAG;
     fi
-  - ./gradlew --console=plain releaseBinaries -PpackageVersion=$BUILD_VERSION
+  - ./gradlew compile -PnativeCompile
```
dubee: Do all of the releases still get built with this native compile change?

houshengbo: no only pick the one for the local env.

mdeuser: are there platform specific environments for testing on? since go allows for native modules, this may become necessary at some point.

Issue #165, opened 2017-11-25
Add OS to User Agent Header


rabbah:
@drcariel commented on [Wed Sep 06 2017](https://github.com/apache/incubator-openwhisk/issues/2704)

Looking for OS information included on UserAgent header as discussed in
https://github.com/apache/incubator-openwhisk/pull/2591

rabbah:
@drcariel is this done?

drcariel:
I believe not - more of a nice to have? or should this be prioritized @rabbah ?

rabbah:
This doesn’t strike me as particularly hard and would provide operational and usage insights so +1 from me.

PR #344, opened 2018-06-19
add OS and CPU architecture to user agent header

drcariel:
Closes https://github.com/apache/incubator-openwhisk-cli/issues/165

Also, should I make a similar change to the client?
https://github.com/apache/incubator-openwhisk-client-go/blob/master/whisk/client.go#L151

@dubee @rabbah

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -21,6 +21,7 @@ import (
 	"errors"
 	"net/http"
 	"os"
+	"runtime"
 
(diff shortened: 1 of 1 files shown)


rabbah:
sounds good re the client. Thanks for doing this.

drcariel:
Updated user agent header looks like so:
`"OpenWhisk-CLI/1.0 (2018-06-19T15:36:55.271-0500) darwin amd64"`

Should be good to go!

Issue #318, opened 2018-06-10
integrate scalafmt, and use it in travis


rabbah:
this repo does not appear to use the upstream scalafmt configuration, nor does it enforce the format.

PR #351, opened 2018-07-09
enable test scala formatting check in Travis

drcariel:
closes https://github.com/apache/incubator-openwhisk-cli/issues/318

Will cause Travis to fail if the scala tests are not formatted correctly. Considering adding documentation around go and scala fmt here as well

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,6 +51,7 @@ install:
 
 before_script:
   - cd $TRAVIS_BUILD_DIR
+  - ./gradlew checkTestScalafmt
   - GO_FILES=$(find . -iname '*.go' -type f | grep -v /vendor/)
   - test -z "$(gofmt -s -l $(echo $GO_FILES))"
   - cd $TRAVIS_BUILD_DIR/..
--- a/.travis.yml
+++ b/.travis.yml
@@ -51,7 +51,7 @@ install:
 
 before_script:
   - cd $TRAVIS_BUILD_DIR
-  - ./gradlew checkTestScalafmt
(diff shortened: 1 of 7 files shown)


drcariel:
@rabbah @dubee let me know if there's anything else you'd like me to add here

drcariel:
Note that in this repo it looks like we actually format the go code if it is not formatted correctly - this is not the case for the client-go repo as I found out the hard way...

rabbah:
@chetanmeh mind reviewing this?

.travis.yml
```diff
@@ -51,6 +51,7 @@ install:
 
 before_script:
   - cd $TRAVIS_BUILD_DIR
+  - ./gradlew checkTestScalafmt
```
chetanmeh: Better to run `checkScalafmtAll` i.e. apply check for all source sets

build.gradle
```diff
@@ -15,10 +15,51 @@
...
+
 plugins {
     id 'com.github.blindpirate.gogradle' version '0.8.1'
 }
 
+subprojects {
+  apply plugin: 'scalafmt'
+  scalafmt.configFilePath = gradle.scalafmt.config
+
+  afterEvaluate {
```
chetanmeh: This should not be required. In core repo its purpose is to customize the launch classpath for controller and invoker apps. For cli builds we do not require that as there is no module with `application` plugin

.travis.yml
chetanmeh: Also use `--console=plain` option

Issue #313, opened 2018-06-07
cannot wsk trigger update


rabbah:
`wsk trigger update x` for any new trigger `x` will fail because the code tries to fetch the trigger first and exists if it's not found. This is not consistent with `wsk action update`, `wsk rule update`, and `wsk package update` all of which will create the entity if it doesn't exist.

PR #353, opened 2018-07-29
Create triggers on update if do not exist

vvraskin:
I wanted to familiarise myself with the cli repo and picked up this bug  https://github.com/apache/incubator-openwhisk-cli/issues/313. 

While working on it, I've realised some spots where unit tests would be welcome, so I added some based on [Ginkgo](https://onsi.github.io/ginkgo/), which I personally find pretty useful for TDD approach, general code coverage and documentation. What do you think guys, is this the right direction to go?
@rabbah @mdeuser @drcariel @houshengbo 

The proposed change still has some nuances to be done properly: gradle deps for ginkgo and gomega,  some duplication and probably a general clean up.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -26,7 +26,7 @@ buildscript {
 }
 
 plugins {
-    id 'com.github.blindpirate.gogradle' version '0.8.1'
+    id 'com.github.blindpirate.gogradle' version '0.10'
 }
 
 subprojects {
@@ -66,8 +66,8 @@ dependencies {
         test name:'github.com/davecgh/go-spew/spew', version:'346938d642f2ec3594ed81d874461961cd0faa76', transitive:false //, tag:'v1.1.0'
         test name:'github.com/pmezard/go-difflib/difflib', version:'792786c7400a136282c1664665ae0a8db921c6c2', transitive:false
 
-        test name:'github.com/onsi/ginkgo'
-        test name:'github.com/onsi/gomega'
+        test name:'github.com/onsi/ginkgo', transitive:false
+        test name:'github.com/onsi/gomega', transitive:false
     }
 }
 
--- a/build.gradle
+++ b/build.gradle
@@ -66,8 +66,13 @@ dependencies {
         test name:'github.com/davecgh/go-spew/spew', version:'346938d642f2ec3594ed81d874461961cd0faa76', transitive:false //, tag:'v1.1.0'
         test name:'github.com/pmezard/go-difflib/difflib', version:'792786c7400a136282c1664665ae0a8db921c6c2', transitive:false
 
-        test name:'github.com/onsi/ginkgo', transitive:false
-        test name:'github.com/onsi/gomega', transitive:false
+        test name:'github.com/onsi/ginkgo', transitive: false
+        test name:'github.com/hpcloud/tail',transitive: false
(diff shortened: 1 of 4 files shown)


rabbah:
Vadim thank you for taking this and adding proper unit test - this will help us improve the cli’s enginnering. I’m all for it.

commands/trigger_test.go
```diff
@@ -0,0 +1,67 @@
+package commands_test
```
dubee: I think this would compile if you change the package to `commands` instead and remove `. "../commands"` from the imports.

build.gradle
```diff
@@ -222,6 +226,12 @@ task goLint(type: com.github.blindpirate.gogradle.Go, dependsOn: getGoLint) {
 
 goCheck.dependsOn(goLint)
 
+task unitTests(type: com.github.blindpirate.gogradle.Go) {
```
dubee: Does this function work with `./gradlew goTest -PgoTags=unit`?

build.gradle
```diff
@@ -56,13 +56,21 @@ dependencies {
...
-        build(['name':'github.com/apache/incubator-openwhisk-client-go/whisk','version':'d7cee96e83a1f38413a1f5286bd524dac72686c9','transitive':false])
+        build(['name':'github.com/apache/incubator-openwhisk-client-go/whisk','version':'2ea476ee7b4a28d96b51fe6ac3488f06adbcaf6e','transitive':false])
         // END - Imported from Godeps
         test name:'github.com/stretchr/testify', version:'b91bfb9ebec76498946beb6af7c0230c7cc7ba6c', transitive:false //, tag: 'v1.2.0'
         test name:'github.com/spf13/viper', version:'aafc9e6bc7b7bb53ddaa75a5ef49a17d6e654be5', transitive:false
         test name:'github.com/cpuguy83/go-md2man/md2man', version:'1d903dcb749992f3741d744c0f8376b4bd7eb3e1', transitive:false //, tag:'v1.0.7'
         test name:'github.com/davecgh/go-spew/spew', version:'346938d642f2ec3594ed81d874461961cd0faa76', transitive:false //, tag:'v1.1.0'
         test name:'github.com/pmezard/go-difflib/difflib', version:'792786c7400a136282c1664665ae0a8db921c6c2', transitive:false
+
+        test name:'github.com/onsi/ginkgo', transitive: false
```
dubee: Can you pin these dependancies to specific commits to avoid random breakages on our end?

commands/trigger.go
```diff
@@ -75,7 +77,7 @@ func feedParameters(feed string, lifecycle string, trigger *QualifiedName, authK
  *
  * NOTE: this function will exit in case of a processing error since it indicates a problem parsing parameters.
  */
-func createOrUpdate(fqn *QualifiedName, trigger *whisk.Trigger, overwrite bool) {
+func createOrUpdate(Client *whisk.Client, fqn *QualifiedName, trigger *whisk.Trigger, overwrite bool) {
```
dubee: Now that `Client` is a parameter, can we delete the comment `// TODO get rid of these global modifiers`?

rabbah: yes, now it's "local".

rabbah:
@vvraskin @dubee let's try to get this in before it rots.

PR #361, opened 2018-08-21
Fix up README

dubee:
Some clean up of the README file here. Also adds Slack and Twitter badges.

PR #362, opened 2018-08-21
Add Govendor as the preferred way of installing dependencies for Go

Juice10:
Adding in a native to Go dependency manager. 
This allows people to not have to go through Gradle if they are doing everything in Go anyway.

Homebrew also requires a native Go dependency manager as opposed to Gradle: https://github.com/Homebrew/homebrew-core/pull/31231

I also looked at [godep](https://github.com/tools/godep) since `wskdeploy` is currently also using that. But according to Godep's own website the project has be Archived and is no longer maintained. This in favor of the many other dependency manager tools.

During my hunt for a Go dependency manager I looked at a number of popular ones including Glide. And eventually I choose govendor because it is very actively being updated and the installation process seemed less involved (`go get ...`) than many other package managers.

### DIFF
--- a/README.md
+++ b/README.md
@@ -18,6 +18,7 @@
 -->
 
 # OpenWhisk Command Line Interface `wsk`
+
 [![License](https://img.shields.io/badge/license-Apache--2.0-blue.svg)](http://www.apache.org/licenses/LICENSE-2.0)
 [![Build Status](https://travis-ci.org/apache/incubator-openwhisk-cli.svg?branch=master)](https://travis-ci.org/apache/incubator-openwhisk-cli)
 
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -36,14 +37,14 @@ The OpenWhisk CLI is written in the Go language. You have two options to build
 the binary locally:
 
 1.  Build using the packaged Gradle scripts (including the 'gogradle' plugin),
-now the preferred build method.
+    now the preferred build method.
```
rabbah: Move up a line?

README.md
```diff
@@ -125,11 +126,12 @@ $ go-bindata -pkg wski18n -o wski18n/i18n_resources.go wski18n/resources
 ```
 
 Unfortunately, it has become necessary to lock dependencies versions to obtain a
-clean build of wsk.  To that end, it's now necessary to populate the `vendors`
```
rabbah: Can you remove unfortunately it has become... seems unnecessary and is a best practice to pin

README.md
Juice10: The line is more than 80 characters so most IDE's cut it off. I indented it so it would align with text above it:
image

README.md
Juice10: Sure

Juice10: Actually, this pull request does that perfectly, maybe we'd like to merge that in to reduce merge conflicts if I also fix it: https://github.com/apache/incubator-openwhisk-cli/pull/361

PR #365, opened 2018-08-28
include test from incubator-openwhisk PR#2941

drcariel:
Test addition after https://github.com/apache/incubator-openwhisk/pull/2941/ was merged

### DIFF
--- a/tests/src/test/scala/org/apache/openwhisk/core/cli/test/ApiGwCliBasicTests.scala
+++ b/tests/src/test/scala/org/apache/openwhisk/core/cli/test/ApiGwCliBasicTests.scala
@@ -1059,7 +1059,7 @@ abstract class ApiGwCliBasicTests extends BaseApiGwTests {
     } finally {
       wsk.action.delete(name = actionName, expectedExitCode = DONTCARE_EXIT)
       apiDelete(basepathOrApiName = testbasepath, expectedExitCode = DONTCARE_EXIT)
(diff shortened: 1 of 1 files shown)


mdeuser:
won't this test go into incubator-openwhisk in the ApiGwRoutemgmtActionTests test suite.. this way both the REST interface is tested.  then the incubator-openwhisk-cli will pick this test up via `ApiGwCliRoutemgmtActionTests extends ApiGwRoutemgmtActionTests`

drcariel:
@mdeuser the test was added to ` incubator-openwhisk/tests/src/test/scala/whisk/core/cli/test/ApiGwRestBasicTests.scala` so I added it here...

If what you say is the "right" way to do this then we need to assess that in a separate PR. Probably another issue even in order to make sure the tests are structured the way you desire.

mdeuser:
some tests are currently structured so that both rest and cli tests have shared tests.  for this particular test, i believe the structure is already in place.  there's a `ApiGwRoutemgmtActionTests.scala` in the incubator-openwhisk that's extended by `ApiGwRestRoutemgmtActionTests.scala` (incubator-openwhisk) for the rest-ful tests as well as `ApiGwCliRoutemgmtActionTests.scala` (incubator-openwhisk-cli) for the cli tests

houshengbo:
@drcariel  please rebase this PR.

drcariel:
@houshengbo done

PR #372, opened 2018-09-20
Add confirm mechanism when delete or update action

ningyougang:
As more and more production system uses openwhisk,
Users will need some feature to protect their action to be
deleted or updated by mistake.


rabbah:
I like the suggestions. Could also store it in whisk properties.

csantanapr:
yep I like the using environment variable.
But I'm not in love with the wording `SAFE`.
How about something along `WSK_CLI_PROMPT_ON_CHANGE` or `WSK_CLI_PROMPT=true|false` (default) to false.
Also this PR only touches on actions.
I think it should be done for all changes (ie. wsk <entity> update | delete)

ningyougang:
@csantanapr , if used `WSK_CLI_PROMPT_ON_CHANGE or WSK_CLI_PROMPT=true|false`
Does this env `WSK_CLI_PROMPT_ON_CHANGE` existed in `openwhisk server side`(not in wsk client side)?

mdeuser:
`WSK_CLI_PROMPT_ON_CHANGE` and/or `WSK_CLI_PROMPT` do not exist in the backend

ningyougang:
@mdeuser 
Regarding `WSK_CLI_PROMPT_ON_CHANGE and/or WSK_CLI_PROMPT do not exist in the backend`
so you mean `WSK_CLI_PROMPT_ON_CHANGE and/or WSK_CLI_PROMPT` exists in `wsk client side`.
if exist in `wsk client side`, for `wsk client users`, they may not know above environment env. 
How to notify `wsk client users` to know above environment env?

rabbah:
A property saved in wskprops would address the environment issue since switching files will then configure the cli as desired. We have no precedent for an env file that changes the cli behavior outside of WSK_CONFIG_FILE so a stand-alone property to me for this behavior seems like the wrong direction.

ningyougang:
@rabbah ,already modified.

rabbah:
On the backend I think the model is finer grained entitlement than we have today, with unix style permissions.

commands/flags.go
```diff
@@ -69,20 +69,22 @@ type FlagsStruct struct {
...
+		promptOnChange    bool
+		cliversion        bool
+		apibuild          bool
+		apibuildno        bool
+		insecure          bool
+		all               bool
+		apihostSet        string
+		apiversionSet     string
+		namespaceSet      string
+		promptOnChangeSet string
```
mdeuser: can this be a boolean flag?  i.e. the present of `--confirm` (typical practice in other clis use `--force`) indicates the desire to override the confirmation.

dubee:
I think a prompt would be better than an additional flag.

ningyougang:
@dubee ,regarding ` a prompt would be better than an additional flag.`
What's mean, can you give a example?

ningyougang:
@rabbah , so we have no need to add the protect mechanism for backend layer, right?

commands/flags.go
ningyougang: Already modified

rabbah:
If by prompt you mean something interactive which requires input on the console then this will make it very inconvenient to actually update a namespace when it is so intended.

ningyougang:
Please don't merge this patch
May be we can solve this problem in backend layer:
https://github.com/apache/incubator-openwhisk/pull/4058

PR #379, opened 2018-11-13
Pin go version used by gogradle

vvraskin:
In some environments where no go binary is installed, gogradle will attempt to download it. If no version is specified, the latest is used. The latest version `1.11` has sharpened the compiler so some code changes in the cli are required to support it, otherwise it won't compile, see https://github.com/apache/incubator-openwhisk-cli/issues/380.

With this change the go version is pinned to the officially supported - `1.9`

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ matrix:
     - os: linux
       sudo: required
       group: deprecated-2017Q3
-      go: 1.9
+      go: 1.9.3
       services: docker
 
 env:
--- a/.travis.yml
+++ b/.travis.yml
@@ -22,7 +22,7 @@ matrix:
     - os: linux
       sudo: required
       group: deprecated-2017Q3
-      go: 1.9.3
+      go: "1.9.3"
       services: docker
 
 env:
--- a/build.gradle
+++ b/build.gradle
@@ -37,6 +37,7 @@ subprojects {
 golang {
     packagePath = 'github.com/apache/incubator-openwhisk-cli' as String
     buildTags = (rootProject.findProperty('goTags')?:'').split(',')
+    goVersion = '1.9.3'
 }
 
 dependencies {


build.gradle
```diff
@@ -37,6 +37,7 @@ subprojects {
 golang {
     packagePath = 'github.com/apache/incubator-openwhisk-cli' as String
     buildTags = (rootProject.findProperty('goTags')?:'').split(',')
+    goVersion = '1.9.3'
```
mdeuser: current go version used to build the published cli is 1.9
https://github.com/apache/incubator-openwhisk-cli/blob/701107f9d73dc455d67e3409a1a9f22235d37bb0/.travis.yml#L25

vvraskin: yeap, but specifying the subversion should be fine as long as the major version doesn't change.

vvraskin: For some reason gogradle cannot download the `1.9.0` on my mac, `1.9.1` onwards are available though

mdeuser: in general, i think we should be consistent when pinning the version of a tool.

vvraskin: can you say which patch version (e.g. 1.9.x) is used by travis?

mdeuser: since travis allows a `1.9.x` specification to allow for the latest patch level, i believe the current travis setting of `1.9` will pull the `1.9.0` version.  this is confirmed by the travis output
```
$ go version
go version go1.9 linux/amd64
```

.travis.yml
```diff
@@ -22,7 +22,7 @@ matrix:
     - os: linux
       sudo: required
       group: deprecated-2017Q3
-      go: 1.9
+      go: 1.9.3
```
mdeuser: oh.. could you please add quotes around the version to conform with a travis/go version best practice ( https://docs.travis-ci.com/user/languages/go/).  thanks!
> Note that, in order to choose Go 1.10, you must use go: "1.10" (a string), not go: 1.10 (a float). Using a float results in the use of Go 1.1.

vvraskin:
@mdeuser do you mind merging it?

PR #382, opened 2018-11-13
Fix distDocker failing on building core images

csantanapr:


csantanapr:
This change decreased the travis build from `38` to `34` minutes

PR #388, opened 2018-11-27
Save Code Associated with Blackbox Actions

dubee:
Currently attempting to save any blackbox action results in an error. These changes update the code saving functionality to  allow for saving of code associated with blackbox actions.

### DIFF
--- a/commands/action.go
+++ b/commands/action.go
@@ -692,7 +692,7 @@ func saveCode(action whisk.Action, filename string) (err error) {
 	exec = *action.Exec
 	runtime = strings.Split(exec.Kind, ":")[0]
 
-	if strings.ToLower(runtime) == BLACKBOX {
+	if strings.ToLower(runtime) == BLACKBOX && exec.Code == nil && *exec.Binary == false {
 		return cannotSaveImageError()
 	} else if strings.ToLower(runtime) == SEQUENCE {
 		return cannotSaveSequenceError()
--- a/tests/src/test/scala/org/apache/openwhisk/core/cli/test/WskCliBasicUsageTests.scala
+++ b/tests/src/test/scala/org/apache/openwhisk/core/cli/test/WskCliBasicUsageTests.scala
@@ -1086,22 +1086,18 @@ class WskCliBasicUsageTests extends TestHelpers with WskTestHelpers {
     val saveName = s"save-as-$name.js"
     val badSaveName = s"bad-directory${File.separator}$saveName"
 
-    Seq(
-      (name, true, false, false),
-      (name, false, true, false),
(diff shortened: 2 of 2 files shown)


commands/action.go
```diff
@@ -692,7 +692,7 @@ func saveCode(action whisk.Action, filename string) (err error) {
 	exec = *action.Exec
 	runtime = strings.Split(exec.Kind, ":")[0]
 
-	if strings.ToLower(runtime) == BLACKBOX {
+	if strings.ToLower(runtime) == BLACKBOX && exec.Code == nil && *exec.Binary == false {
```
mdeuser: what are the kinds of blackbox action in the following scenarios:
- exec.Code == nil and exec.Binary ==  `true`
- exec.Code == nil and exec.Binary ==  `false`

will exec.Binary always have a value - that is, it can never be `nil` ?

dubee: Yes, the binary field should always have a value. @mdeuser

mdeuser: @dubee - and the kind of actions represented by the following states?
- exec.Code == nil and exec.Binary == true
- exec.Code == nil and exec.Binary == false

dubee: Blackbox actions.

dubee: `exec.Code == nil and exec.Binary == true` would be an action that has a zip file as code or a Java action. `exec.Code == nil and exec.Binary == false` represents a blackbox action without any code provided when the action was created. @mdeuser

dubee: When creating a blackbox action, code can be added to the action. For documentation see https://github.com/apache/incubator-openwhisk/blob/master/docs/actions-docker.md#creating-native-actions.

Here's a example of creating a blackbox action with additional initialization code: `wsk action create actionName code.zip --docker dockerImage`. @mdeuser

dubee:
@mdeuser, anything else on this one?

commands/action.go
mdeuser: thanks.. that clarified it for me. the *save* operation will write out some binary file.  the exact specification of that binary is up to the developer to know and treat it accordingly.

i suppose the `else` just following this check is really not needed..

dubee: @mdeuser, do you mean `else if strings.ToLower(runtime) == SEQUENCE`?

mdeuser: yeah.. thinking this is equivalent
```
	if strings.ToLower(runtime) == BLACKBOX && exec.Code == nil && *exec.Binary == false {
		return cannotSaveImageError()
	}

        if strings.ToLower(runtime) == SEQUENCE {
		return cannotSaveSequenceError()
        }
```

PR #389, opened 2018-11-30
Add CA certificate processing.

jonpspri:
Create a '--cacert' flag that allows wsk to honor self-signing CAs if the certificate is provided. While not eliminating the need for wsk -i this does provide a more secure alternative for environments where the signing authority is stable, not needed in the system setup, and not easily changed (e.g. Kubernetes cluster development/integration installs).

https://github.com/apache/incubator-openwhisk-client-go/pull/109 should be committed prior to this CR.

I have an Apache CLA on file.


jonpspri:
Note the current Travis failure is because apache/incubator-openwhisk-client-go#109 has not yet been applied.  It will be necessary to restart Travis once that PR has been applied.

jonpspri:
Closing as stale.

PR #395, opened 2018-12-13
Update to the most recent version of gradle

2rob:
gradle builds fail with: "Could not determine java version from '11.0.1'." unless its using an updated version of gradle.  was successful with gradle 5.0


rabbah:
Thanks @2rob - this is OK for this repo but might cause issues across all the openwhisk repos which are not on v5. I'm guessing you don't have java 8 or 10?

Issue #390, opened 2018-12-04
wsk activation poll does not update the since time


chetanmeh:
`wsk activation poll` when invoked without any explicit time parameters performs a query like 

```
/api/v1/namespaces/_/activations?docs=true&limit=0&since=1542422386538&skip=0
```

Where it fetches full docs without any explicit limit and `since` set to epoch time at the start of poll. At times we have seen queries where the difference between current time and `since` is in days

```
[2018-12-04T15:15:10.467Z] [INFO] [#tid_1rrQBHlr8f8Xomky3qi3fsELzTIpH7CB] GET /api/v1/namespaces/_/activations docs=true&limit=0&since=1543338217244&skip=0
```

Here _1543338217244_ is ~ Nov-27. This happens because some developers tend to leave poll command running in some forgotten console tab. Due to this each such poll command tries to fetch same set of 200 records (due to limit=0) over and over again putting load on db

chetanmeh:
From looking at [code][1] it appears that `pollSince` does not get updated once the loop starts. 

@rabbah suggested on slack that 

> Since is just where to start (instead of “now” you start in the past and then update the cursor so that since moves up to the last stamp in a given query)

[1]: https://github.com/apache/incubator-openwhisk-cli/blob/master/commands/activation.go#L356

PR #396, opened 2018-12-14
Update the pollSince time based on latest activation start time

chetanmeh:
Change the `pollSince` time for each loop by basing it on the latest `Start` time seen in activation result

See #390 for background details

### Testing Approach

While the fix is small it was tricky to validate the change. This PR add a test HTTP rever proxy which is used to intercept the calls between the cli client and the controller. The test then intercepts the request to check if since time gets changed or not

### DIFF
--- a/commands/activation.go
+++ b/commands/activation.go
@@ -387,6 +387,9 @@ var activationPollCmd = &cobra.Command{
 					printJSON(activation.Logs)
 					reported[activation.ActivationID] = true
 				}
+				if activation.Start > pollSince {
+					pollSince = activation.Start
+				}
 			}
 			time.Sleep(time.Second * 2)
 		}
--- a/tests/src/test/scala/system/basic/HttpProxy.scala
+++ b/tests/src/test/scala/system/basic/HttpProxy.scala
@@ -17,14 +17,16 @@
 package system.basic
 import java.net.ServerSocket
 
-import akka.http.scaladsl.Http
+import akka.http.scaladsl.{Http, HttpsConnectionContext}
 import akka.http.scaladsl.model.{HttpRequest, HttpResponse, Uri}
 import akka.http.scaladsl.model.Uri.Authority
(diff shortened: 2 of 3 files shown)


chetanmeh:
@csantanapr @rabbah @dubee Please review

dariosm:
Hi, about this change: https://github.com/apache/incubator-openwhisk-cli/pull/396/files#diff-ad777399e121d471d84d4fd5bf456bceR390
@chetanmeh I believe it affects nested cloud functions. I’m experiencing activation logs missing while running `activation poll` and invoking a cloud function that triggers others. Actually, I can see inner cloud functions logs but never the logs of the main function than executed them.
We’ve discussed this with @mdeuser (thanks him for pointing me to this PR). I’ve reported the issue before here: https://ibm-cloud-tech.slack.com/team/UGVBSBWGH)

mdeuser:
the activation polling scenario in question is:
1. activation A starts and continues to run
2. activation B starts and quickly completes
3. activation poll reports B completed and updates the `since` time
4. activation A completes
5. activation poll does not report activation A as being completed.
i believe the GET /activations api's `since` parameter is based off the start time, not the end time...  but @rabbah will know for sure

rabbah:
It’s possible to miss activations because there’s no guarantee the activations are committed to the datastore in order. The cursor is the cli advances based on the last fetched activations. The polling doesn’t back fill.

mdeuser:
@rabbah - thanks fr the input.  does the backend use the activation start time or end time when processing the activation list `since` parameter?

while not a guarantee, i'm thinking the activation order will likely be maintained most of the time - especially for activations completing in sequence and seconds apart.  a polled-for, long running sequence action will probably run into this same issue.

rabbah:
The activations in the database are indexed using the activation start time. The controller queries the database using the `start` key equal to the `since` parameter, and the `end` key equal to the `upto` parameter.

PR #401, opened 2019-01-26
Recognize balx files as binary.

rabbah:


rabbah:
it would be a binary - how do you deal with jar files today for java? do you save them?
we could just save the balx file.

dubee:
Currently we just save the `.jar` file for Java actions.

Issue #398, opened 2019-01-18
Failed to go build -o wsk using README.md instructions


nitishm:
I cannot get the wsk CLI to build, following the README.md instructions page.

The `go build -o wsk` command fails with the following error -
```
go build -o wsk
# github.com/apache/incubator-openwhisk-cli/commands
commands/action.go:911:10: limits.Concurrency undefined (type *"github.com/apache/incubator-openwhisk-cli/vendor/github.com/apache/incubator-openwhisk-client-go/whisk".Limits has no field or method Concurrency)
commands/project.go:40:20: cannot use cobraCMD (type *"github.com/apache/incubator-openwhisk-cli/vendor/github.com/spf13/cobra".Command) as type *"github.com/spf13/cobra".Command in argument to cmd.Deploy
commands/project.go:51:22: cannot use cobraCMD (type *"github.com/apache/incubator-openwhisk-cli/vendor/github.com/spf13/cobra".Command) as type *"github.com/spf13/cobra".Command in argument to cmd.Undeploy
commands/project.go:63:20: cannot use cobraCMD (type *"github.com/apache/incubator-openwhisk-cli/vendor/github.com/spf13/cobra".Command) as type *"github.com/spf13/cobra".Command in argument to cmd.Deploy
commands/project.go:74:26: cannot use cobraCMD (type *"github.com/apache/incubator-openwhisk-cli/vendor/github.com/spf13/cobra".Command) as type *"github.com/spf13/cobra".Command in argument to cmd.ExportCmdImp
```

I can resolve the `cobra` errors by deleting the package from the vendor dir (not ideal), but the other issue,

```
go build -o wsk
# github.com/apache/incubator-openwhisk-cli/commands
commands/action.go:911:10: limits.Concurrency undefined (type *"github.com/apache/incubator-openwhisk-cli/vendor/github.com/apache/incubator-openwhisk-client-go/whisk".Limits has no field or method Concurrency)
```
, seems to be stemming from a mismatched client-go version. 

I can see the master branch for github.com/apache/incubator-openwhisk-client-go/ now has the `Concurrency` field added to the `Limits` struct (https://github.com/apache/incubator-openwhisk-client-go/blob/master/whisk/shared.go#L100), but `go get` seems to fetch an older commit of the client-go library causing the build to fail.

dubee:
@nitishm, have you tried `./gradlew compile -PnativeCompile`?

nitishm:
[...]

PR #402, opened 2019-01-26
Update README and add action unit test

rabbah:
I documented a workaround because of #398. I will post more details in the issue. 

I also added a test case for my previous PR #401 and docs for how to run unit tests with go.

### DIFF
--- a/commands/action_test.go
+++ b/commands/action_test.go
@@ -1,0 +1,51 @@
+// +build unit
+
+/*
+ * Licensed to the Apache Software Foundation (ASF) under one or more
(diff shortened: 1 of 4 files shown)


commands/action_test.go
```diff
@@ -0,0 +1,51 @@
...
+package commands
+
+import (
+	"encoding/base64"
+	"github.com/stretchr/testify/assert"
+	"os"
+	"testing"
+)
+
+func TestBallerinaBinaryFile(t *testing.T) {
```
csantanapr: Very nice 👍🏻 to have Go tests

PR #404, opened 2019-01-30
corrected description of go build steps

matthiaskubik:
If you start off a clean GOPATH tree, it'll fail at `go get https://github.com/apache/incubator-openwhisk-cli` as prereqs are missing that are installed later. Need to run a `git clone` instead.

### DIFF
--- a/README.md
+++ b/README.md
@@ -114,8 +114,9 @@ Then download the source code of the OpenWhisk CLI and the dependencies by
 typing:
 
 ```
-$ cd $GOPATH
-$ go get github.com/apache/incubator-openwhisk-cli
+$ mkdir -p $GOPATH/github.com/apache
+$ cd $GOPATH/github.com/apache
+$ git clone https://github.com/apache/incubator-openwhisk-cli
 $ cd $GOPATH/src/github.com/apache/incubator-openwhisk-cli
 ```
 


dubee:
@mdeuser, can you verify this one?

README.md
```diff
@@ -114,8 +114,9 @@ Then download the source code of the OpenWhisk CLI and the dependencies by
 typing:
 
 ```
-$ cd $GOPATH
-$ go get github.com/apache/incubator-openwhisk-cli
+$ mkdir -p $GOPATH/github.com/apache
+$ cd $GOPATH/github.com/apache
+$ git clone https://github.com/apache/incubator-openwhisk-cli
```
mdeuser: i recommend retaining the "go" flavor of this installation, so instead of the `git` command, how about just using the `-d` (download only; no compile/install) option.
```
go get -d github.com/apache/incubator-openwhisk-cli
```
this should avoid the
```
src\github.com\apache\incubator-openwhisk-cli\wski18n\i18n.go:98:16: undefined: Asset
```
error seen without the `-d` option

mrutkows:
@matthiaskubik can you add the instructions for "govendor" that causes both the SHA1 and the UTC timestamp to be updated in vendor.json (as well as update the build.gradle and how they relate)?

mrutkows:
Closing. The project no longer uses govendor and with gomod no longer requires "getting" or cloning into a specific GOPATH,

Issue #403, opened 2019-01-29
save balx file with --save


rabbah:
from https://github.com/apache/incubator-openwhisk-cli/pull/401 @dubee pointed out we should save the balx files.

PR #405, opened 2019-01-31
Save Ballerina Binaries

dubee:
Save Ballerina binaries with the proper extension.

Closes https://github.com/apache/incubator-openwhisk-cli/issues/403


rabbah:
ok that was easy --- do you mind adding a go unit test https://github.com/apache/incubator-openwhisk-cli/blob/master/commands/action_test.go if it doesn't end up requiring a lot of refactoring?

csantanapr:
FTW ! Go unit tests

PR #418, opened 2019-03-05
update versions of wskdeploy and client-go

dgrove-oss:


dgrove-oss:
Going to do this in multiple baby steps; closing in favor of #421 (and later separate PR to update client-go).

PR #422, opened 2019-03-11
Update client-go version to af04d34b (02/12/2019)

dgrove-oss:


dgrove-oss:
Doing a binary search to identify which change to client-go is causing the failure in WskCliConsoleTests.  (Updating all the way to client-go version 4286a82 reliably fails).

dgrove-oss:
Closing so we can do it in 1 step

PR #423, opened 2019-03-12
Attempt to upgrade to client-go version 67f06bc

dgrove-oss:


dgrove-oss:
I expect this PR to fail (vs. #422 which succeed). If this happens, it will identify the problematic commit in client-go that is blocking upgrading the cli to latest.

dgrove-oss:
closing so we can do the bump in one step

PR #433, opened 2019-04-10
adapt tests for provide-api-key annotation upstream

matthiaskubik:
- bumped gradle version due to SSL errors
- copied changes from https://github.com/apache/incubator-openwhisk/pull/4423

### DIFF
--- a/tests/src/test/scala/org/apache/openwhisk/core/cli/test/WskCliBasicUsageTests.scala
+++ b/tests/src/test/scala/org/apache/openwhisk/core/cli/test/WskCliBasicUsageTests.scala
@@ -632,7 +632,9 @@ class WskCliBasicUsageTests extends TestHelpers with WskTestHelpers {
           baseAnnotations ++ Parameters(WhiskAction.provideApiKeyAnnotationName, JsFalse)
         } else baseAnnotations
 
-        action.getFieldJsValue("annotations").convertTo[Set[JsObject]] shouldBe testAnnotations.toJsArray
+        removeCLIHeader(action.stdout).parseJson.asJsObject
+          .fields("annotations")
+          .convertTo[Set[JsObject]] shouldBe testAnnotations.toJsArray
           .convertTo[Set[JsObject]]
       }
   }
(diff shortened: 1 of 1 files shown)


gradle/wrapper/gradle-wrapper.properties
```diff
@@ -5,4 +5,4 @@ distributionBase=GRADLE_USER_HOME
 distributionPath=wrapper/dists
 zipStoreBase=GRADLE_USER_HOME
 zipStorePath=wrapper/dists
-distributionUrl=https\://services.gradle.org/distributions/gradle-4.1-bin.zip
```
rabbah: Is this necessary? Might change required java version. I have to check but I think we use the same version in all the repos still.

rabbah: Let me rephrase - if we’re doing it here, I’m assuming we will have to do it across all the repos.

gradle/wrapper/gradle-wrapper.properties
```diff
@@ -5,4 +5,4 @@ distributionBase=GRADLE_USER_HOME
...
+distributionUrl=https\://services.gradle.org/distributions/gradle-4.10.2-bin.zip
```
rabbah: With the change here do we have to make the same change across all the repos? I don’t know how many are on the old version. I’m assuming they will suffer the same ssl error, no?

matthiaskubik: I'd be in favor of it but have not too much insight of what might break, so I rely on your advise...

rabbah:
This PR should be merged _without_ squashing independent commits in case we need to roll back. Ideally orthogonal changes are in different PRs.

gradle/wrapper/gradle-wrapper.properties
rabbah: https://github.com/apache/incubator-openwhisk/pull/4111/files#diff-fd9d61e37d939c941e7a62077f38d62dR6 the openwhisk repo is already on this version.

rabbah:
@matthiaskubik I'm going to bump the gradle version to 4.10.2 in this PR https://github.com/apache/incubator-openwhisk-cli/pull/395. This PR then will address just the test changes. Sorry for the confusion. I'm trying to reduce the work for you.

rabbah:
@matthiaskubik i rebased your PR to master to pick up the gradle bump and retitled your PR accordingly. Thanks for the contribution and fixing the tests.

rabbah:
Thanks @matthiaskubik for the contribution.

PR #434, opened 2019-04-25
remove namespace property from wskprops

### LINKED ISSUE #137: remove NAMESPACE from wskprops file and wsk property subcommand
@sjfink commented on [Mon Apr 17 2017](https://github.com/apache/incubator-openwhisk/issues/2150)

```
$ wsk property get --namespace
whisk namespace	      _
```

I propose `wsk property get --namespace` should always return the explicit name of the default namespace, and not `_` or `default`.

rabbah:
The setting of a namespace in wskprops can lead to surprising failures when it does not match the auth key. This value used to be there when namespaces and keys were not 1-1 but that hasn't been the case for a long time.

This patch remove --namespace from property set/unset but preserves --namespace in property get.

Partially fixes #137. Will open a subsequent PR to re-purpose `wsk namespace get`.

### DIFF
--- a/commands/action.go
+++ b/commands/action.go
@@ -939,7 +939,7 @@ func nonNestedError(errorMessage string) error {
 }
 
 func actionParseError(cmd *cobra.Command, args []string, err error) error {
-	whisk.Debug(whisk.DbgError, "parseAction(%s, %s) error: %s\n", cmd, args, err)
+	whisk.Debug(whisk.DbgError, "parseAction(%s, %s) error: %s\n", cmd.Name(), args, err)
 
 	errMsg := wski18n.T(
 		"Invalid argument(s). {{.required}}",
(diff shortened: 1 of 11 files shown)


neerajmangal:
sorry @rabbah, for the late response. I tested the changes and all looks good.

sciabarracom:
I built the cli and I tried with an existing configuration with a namespace:

Old: 
wsk -i action list
error: Unable to obtain the list of actions for namespace 'neuronix_cloud': The supplied authentication is not authorized to access 'neuronix_cloud'. (code nVQD2dAZZbjgN4Z1fqsXkE7Q8aAFVFSi)
Run 'wsk --help' for usage.

New:
./wsk -i action list
actions

rabbah:
Thanks @neerajmangal @sciabarracom.

devaughnsaxa:
@rabbah I think this doc needs update then
https://openwhisk.apache.org/documentation.html#wsk-cli

It still shows the --namespace flag

PR #437, opened 2019-04-30
Fix tests when whisk.feature.requireApiKeyAnnotation=false

matthiaskubik:
Tests are failing if whisk.feature.requireApiKeyAnnotation=false


tests/src/test/scala/org/apache/openwhisk/core/cli/test/WskCliBasicUsageTests.scala
```diff
@@ -659,18 +657,29 @@ class WskCliBasicUsageTests extends TestHelpers with WskTestHelpers {
     }
 
     wsk.action.create(name, file, web = Some("true"), update = true)
+    val expectedExistingAnnotations = if (requireAPIKeyAnnotation) {
```
mdeuser: instead of having duplicate JsArray entry lines (common entries are coded twice), how about defining a single "base" JsArray and conditionally adding the array element
```
JsObject("key" -> JsString(WhiskAction.provideApiKeyAnnotationName), "value" -> JsBoolean(false)),
```

PR #442, opened 2019-06-13
Update all files using Mini license header to use full ASF header

mrutkows:


commands/action.go
```diff
@@ -40,7 +40,7 @@ import (
 
 const (
 	MEMORY_LIMIT      = 256
-	TIMEOUT_LIMIT     = 60000
+	TIMEOUT_LIMIT     = 600000
```
rabbah: slipped in?

more
```diff
@@ -0,0 +1,13 @@
+./commands/property.go:		// get [1;36;40m[Kcurrent[m[K props
```
rabbah: delete this file?

commands/action.go
mrutkows: was playing with a month or so ago...  likely on my master... good catch

more
mrutkows: all better... was seeing what would happen if client limit value exceeded server-side...

PR #444, opened 2019-06-26
Allow log stripping to tolerate a missing stream identifier.

### LINKED ISSUE #443: go unit tests are not running in travis

rabbah:
This PR allows log stripping to tolerate a missing log stream identifier. With the standalone controller, the `stdout` or `stderr` stream identifier is not readily available. This defeats the current regex matcher and logs are not stripped as a result. The patch corrects this.

Also closes #443.


rabbah:
This PR should fail because a unit test fails. A second commit will correct this but I want to address #443 first and will do so as part of this PR.

commands/util.go
```diff
@@ -273,11 +273,15 @@ func makeDefaultHeader(collection interface{}) string {
 }
 
 func stripTimestamp(log string) (strippedLog string) {
-	regex := regexp.MustCompile("[a-zA-Z0-9\\s]*(stdout|stderr):\\s(.*)")
+	// parses out the timestamp if it exists first
+	// the timestamp expected format is YYYY-MM-DDTHH:MM:SS.[0-9]+Z
+	// an optional " stdout" or " stderr" stream identifier
+	// and the rest as the log line
+    regex := regexp.MustCompile("\\d{4}-[01]{1}\\d{1}-[0-3]{1}\\d{1}T[0-2]{1}\\d{1}:[0-6]{1}\\d{1}:[0-6]{1}\\d{1}.\\d+Z( (stdout|stderr):)?\\s(.*)")
```
rabbah: this regex does make a semantic change.
previously, the following line

`anything stdout: this is stdout` 
would be stripped to 
`this is stdout`

but with this patch the line will remain intact.

rabbah:
Add unit tests to travis.

PR #449, opened 2019-07-02
update wskdeploy and client-go dependencies

dgrove-oss:


rabbah:
I'd be in favor of separately dropping gradle for building the cli... it is _far too slow_ compared to go.

dgrove-oss:
I'd love to see someone rationalize our go dependency/build story across client-go, wskdeploy, and cli projects.  We have multiple mechanisms and poor documentation of what is actually the preferred way to do it.

rabbah:
yes - i'd use `go` and vendoring for all of these. there's a nasty circularity with wskdeploy i didn't figure out how to remove (actually wonder if wskdeploy should be merged into this repo, what's the value of it being a separate project but that's also a different topic of discussion for the dev list).

PR #457, opened 2019-10-11
gradle version 5

dgrove-oss:


rabbah:
bad repo config.

dubee:
@dgrove-oss, @dgrove-oss, this was merged without a passing Travis build?

rabbah:
Travis is broken because of pip issue which affected the main repo, this and likely others. 

The fix is in this pr https://github.com/apache/openwhisk-cli/pull/458

dubee:
@rabbah, we are seeing build timeouts due to this update since tests suites are running multiple times now. It appears `-Dtest.single` was removed from Gradle 5 resulting in test suites running multiple times. From https://docs.gradle.org/current/userguide/upgrading_version_4.html: `The -Dtest.single command-line option has been removed — use test filtering instead.`

dubee:
@rabbah, Travis for this repo appears to be using `-Dtest.single` as well.

rabbah:
👍

 my fault for merging while Travis was broke.   

Do you have a fix already to open a pr? Or do you propose to revert the pr? Either way we’ll need to fix for gradle 5 since I think all the repos have migrated or are migrating. Are there any other repos that are affected by the single parameter that you’ve found?

The repo config as I noted is broken as it shouldn’t have allowed me to merge the PR. I already made a list of all the repos that are too permissive to address via apache/jira. 

In any case Travis is still broken and we need the patches Chetan started.

dgrove-oss:
piling on changes to #458 in #460...

rabbah:
Thanks Dave.

PR #475, opened 2020-01-04
convert secretValue int64 to string type

steven0711dong:
Manual test included all the possible value combination of --web-secure and --annotation flags. The following scenarios of combination were tested: 

1. --web-secure true -a require-whisk-auth: integer
2. --web-secure true -a require-whisk-auth: string 
3. --web-secure false -a require-whisk-auth: integer 
4. --web-secure false -a require-whisk-auth: string
5. --web-secure true
6. --web-secure false
7. -a require-whisk-auth integer
8. -a require-whisk-auth string

close https://github.ibm.com/BlueMix-Fabric/bluemix-openwhisk-cli/issues/329

### DIFF
--- a/commands/action.go
+++ b/commands/action.go
@@ -27,6 +27,7 @@ import (
 	"os"
 	"path/filepath"
 	"reflect"
+	"strconv"
 	"strings"
 	"time"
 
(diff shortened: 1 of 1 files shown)


steven0711dong:
Some of the logic associated with flag --web-secure relies on the type of the require-whisk-auth value. Converting number to string changes a lot of things. I'm still in the middle of testing out everything.

PR #477, opened 2020-01-07
build binary test artifacts

dgrove-oss:


dgrove-oss:
@mdeuser @steven0711dong -- need to run the task `tests:buildArtifacts` to generate the binary artifacts

mdeuser:
thanks a lot @dgrove-oss - is the `helloJava.jar` expected to be generated by that gradle target?

dgrove-oss:
The Java artifact will only be generated if there is a Java8 JVM available.  Looking at the log, it appears that the JVM on the test VM was Java11 so the jar generation was skipped.

mdeuser:
ugh.. https://github.com/apache/openwhisk-cli/blob/368da5bf3d34b4a9b64204a0876618d79e014146/tests/src/test/scala/org/apache/openwhisk/core/cli/test/WskConfigTests.scala#L244

we're in a new decade!  i believe changing `201.*` to `20.*` should fix one of the failing tests

PR #479, opened 2020-01-21
Trigger parameter issue

steven0711dong:
close https://github.com/apache/openwhisk-cli/issues/290, https://github.com/apache/openwhisk-cli/issues/293

### DIFF
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -19,6 +19,7 @@ package commands
 
 import (
 	"errors"
+	"fmt"
 	"net/http"
 	"os"
 	"runtime"
@@ -131,9 +132,11 @@ func getValueFromArgs(args []string, argIndex int, parsedArgs []string) ([]strin
 	return parsedArgs, args, whiskErr
 }
 
-func parseArgs(args []string) ([]string, []string, []string, error) {
+func parseArgs(args []string) ([]string, []string, []string, []string, []string, error) {
 	var paramArgs []string
 	var annotArgs []string
+	var feedParamArgs []string
+	var triggerParamArgs []string
 	var whiskErr error
 
 	i := 0
@@ -147,14 +150,14 @@ func parseArgs(args []string) ([]string, []string, []string, error) {
 					map[string]interface{}{"err": whiskErr})
 				whiskErr = whisk.MakeWskError(errors.New(errMsg), whisk.EXIT_CODE_ERR_GENERAL, whisk.DISPLAY_MSG,
 					whisk.DISPLAY_USAGE)
-				return nil, nil, nil, whiskErr
+				return nil, nil, nil, nil, nil, whiskErr
 			}
 
 			filename := paramArgs[len(paramArgs)-1]
 			paramArgs[len(paramArgs)-1], whiskErr = ReadFile(filename)
 			if whiskErr != nil {
 				whisk.Debug(whisk.DbgError, "readFile(%s) error: %s\n", filename, whiskErr)
-				return nil, nil, nil, whiskErr
+				return nil, nil, nil, nil, nil, whiskErr
 			}
 		} else if args[i] == "-A" || args[i] == "--annotation-file" {
 			annotArgs, args, whiskErr = getValueFromArgs(args, i, annotArgs)
@@ -164,14 +167,14 @@ func parseArgs(args []string) ([]string, []string, []string, error) {
 					map[string]interface{}{"err": whiskErr})
 				whiskErr = whisk.MakeWskError(errors.New(errMsg), whisk.EXIT_CODE_ERR_GENERAL, whisk.DISPLAY_MSG,
 					whisk.DISPLAY_USAGE)
-				return nil, nil, nil, whiskErr
+				return nil, nil, nil, nil, nil, whiskErr
 			}
 
 			filename := annotArgs[len(annotArgs)-1]
 			annotArgs[len(annotArgs)-1], whiskErr = ReadFile(filename)
 			if whiskErr != nil {
 				whisk.Debug(whisk.DbgError, "readFile(%s) error: %s\n", filename, whiskErr)
-				return nil, nil, nil, whiskErr
+				return nil, nil, nil, nil, nil, whiskErr
 			}
 		} else if args[i] == "-p" || args[i] == "--param" {
 			paramArgs, args, whiskErr = getKeyValueArgs(args, i, paramArgs)
@@ -181,9 +184,10 @@ func parseArgs(args []string) ([]string, []string, []string, error) {
 					map[string]interface{}{"err": whiskErr})
 				whiskErr = whisk.MakeWskError(errors.New(errMsg), whisk.EXIT_CODE_ERR_GENERAL, whisk.DISPLAY_MSG,
 					whisk.DISPLAY_USAGE)
-				return nil, nil, nil, whiskErr
+				return nil, nil, nil, nil, nil, whiskErr
 			}
 		} else if args[i] == "-a" || args[i] == "--annotation" {
+			fmt.Println("4")
 			annotArgs, args, whiskErr = getKeyValueArgs(args, i, annotArgs)
 			if whiskErr != nil {
 				whisk.Debug(whisk.DbgError, "getKeyValueArgs(%#v, %d) failed: %s\n", args, i, whiskErr)
@@ -191,7 +195,23 @@ func parseArgs(args []string) ([]string, []string, []string, error) {
 					map[string]interface{}{"err": whiskErr})
 				whiskErr = whisk.MakeWskError(errors.New(errMsg), whisk.EXIT_CODE_ERR_GENERAL, whisk.DISPLAY_MSG,
 					whisk.DISPLAY_USAGE)
-				return nil, nil, nil, whiskErr
+				return nil, nil, nil, nil, nil, whiskErr
+			}
+		} else if args[i] == "-F" || args[i] == "--feed-param" {
+			feedParamArgs, args, whiskErr = getKeyValueArgs(args, i, feedParamArgs)
+			if whiskErr != nil {
+				whisk.Debug(whisk.DbgError, "getKeyValueArgs(%#v, %d) failed: %s\n", args, i, whiskErr)
+				whiskErr = whisk.MakeWskError(whiskErr, whisk.EXIT_CODE_ERR_GENERAL, whisk.DISPLAY_MSG,
+					whisk.DISPLAY_USAGE)
+				return nil, nil, nil, nil, nil, whiskErr
+			}
+		} else if args[i] == "-T" || args[i] == "--trigger-param" {
+			triggerParamArgs, args, whiskErr = getKeyValueArgs(args, i, triggerParamArgs)
+			if whiskErr != nil {
+				whisk.Debug(whisk.DbgError, "getKeyValueArgs(%#v, %d) failed: %s\n", args, i, whiskErr)
+				whiskErr = whisk.MakeWskError(whiskErr, whisk.EXIT_CODE_ERR_GENERAL, whisk.DISPLAY_MSG,
+					whisk.DISPLAY_USAGE)
+				return nil, nil, nil, nil, nil, whiskErr
 			}
 		} else {
 			i++
@@ -200,16 +220,18 @@ func parseArgs(args []string) ([]string, []string, []string, error) {
 
 	whisk.Debug(whisk.DbgInfo, "Found param args '%s'.\n", paramArgs)
 	whisk.Debug(whisk.DbgInfo, "Found annotations args '%s'.\n", annotArgs)
+	whisk.Debug(whisk.DbgInfo, "Found feed param args '%s'.\n", feedParamArgs)
+	whisk.Debug(whisk.DbgInfo, "Found trigger param args '%s'.\n", triggerParamArgs)
 	whisk.Debug(whisk.DbgInfo, "Arguments with param args removed '%s'.\n", args)
 
-	return args, paramArgs, annotArgs, nil
+	return args, paramArgs, annotArgs, feedParamArgs, triggerParamArgs, nil
 }
 
 func Execute() error {
 	var err error
 
 	whisk.Debug(whisk.DbgInfo, "wsk args: %#v\n", os.Args)
-	os.Args, Flags.common.param, Flags.common.annotation, err = parseArgs(os.Args)
+	os.Args, Flags.common.param, Flags.common.annotation, Flags.trigger.feedParam, Flags.trigger.triggerParam, err = parseArgs(os.Args)
 
 	if err != nil {
 		whisk.Debug(whisk.DbgError, "parseParams(%s) failed: %s\n", os.Args, err)
--- a/commands/commands.go
+++ b/commands/commands.go
@@ -19,7 +19,6 @@ package commands
 
 import (
 	"errors"
-	"fmt"
 	"net/http"
 	"os"
 	"runtime"
@@ -187,7 +186,6 @@ func parseArgs(args []string) ([]string, []string, []string, []string, []string,
 				return nil, nil, nil, nil, nil, whiskErr
 			}
 		} else if args[i] == "-a" || args[i] == "--annotation" {
-			fmt.Println("4")
 			annotArgs, args, whiskErr = getKeyValueArgs(args, i, annotArgs)
 			if whiskErr != nil {
 				whisk.Debug(whisk.DbgError, "getKeyValueArgs(%#v, %d) failed: %s\n", args, i, whiskErr)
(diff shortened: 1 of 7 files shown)


tests/src/test/scala/system/basic/WskCliBasicTests.scala
```diff
@@ -451,7 +451,7 @@ class WskCliBasicTests extends TestHelpers with WskTestHelpers {
     getJSONFromResponse(trigger.stdout, true).fields("parameters") shouldBe JsArray(
       JsObject("key" -> JsString("a"), "value" -> JsString("A")))
     getJSONFromResponse(trigger.stdout, true).fields("publish") shouldBe false.toJson
-    getJSONFromResponse(trigger.stdout, true).fields("version") shouldBe "0.0.2".toJson
+    getJSONFromResponse(trigger.stdout, true).fields("version") shouldBe "0.0.1".toJson
```
rabbah: this is semantic changing - it suggests there was a call to trigger update somewhere before that is no longer occurring.

in fact on line 438 and 439 you see there was a `create` and then an `update`. So I think this change is not correct and should be reverted.

commands/trigger.go
```diff
@@ -594,10 +579,42 @@ func (t *Trigger) Update(Client *whisk.Client, args []string) error {
 			}
 		}
 
+		//if there is no feed attached to this trigger
+		if len(fullFeedName) < 1 {
+			//but user indicate feed parameter change, we issue error message.
+			if len(Flags.trigger.feedParam) > 0 {
+				whisk.Debug(whisk.DbgError, "trigger %+v is found but it does not contain a feed. \n", trigger)
+				errStr := wski18n.T("triggerDoesNotContainFeed", map[string]interface{}{"name": qualifiedName.GetEntityName()})
```
rabbah: Please look at using a constant in https://github.com/apache/openwhisk-cli/blob/master/commands/messages.go instead of a string `triggerDoesNotContainFeed` which requires a matching entry in the localization file.

steven0711dong: `triggerDoesNotContainFeed` represents a very specific message that tells users they've run into a specific situation that will results to error. This happens when they tries to update feed params when the trigger contains no feed. I cannot find any constant in messages.go file that represents this. There is only one constant that's relevant, which is `FEED_CONFIGURATION_FAILURE`, but this is too broad and general error message.

commands/commands.go
```diff
@@ -191,7 +192,17 @@ func parseArgs(args []string) ([]string, []string, []string, error) {
...
 				whiskErr = whisk.MakeWskError(errors.New(errMsg), whisk.EXIT_CODE_ERR_GENERAL, whisk.DISPLAY_MSG,
 					whisk.DISPLAY_USAGE)
-				return nil, nil, nil, whiskErr
+				return nil, nil, nil, nil, whiskErr
+			}
+		} else if args[i] == "-fp" || args[i] == "--feed-param" {
+			feedParamArgs, args, whiskErr = getKeyValueArgs(args, i, feedParamArgs)
+			if whiskErr != nil {
+				whisk.Debug(whisk.DbgError, "getKeyValueArgs(%#v, %d) failed: %s\n", args, i, whiskErr)
+				errMsg := wski18n.T("The parameter arguments are invalid: {{.err}}",
```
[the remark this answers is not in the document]
steven0711dong: I realize that it is sufficient to rely on error message produced by getKeyValueArgs. So I removed the unnecessary error string.

commands/trigger.go
steven0711dong: You are correct. I do not need to add another string in the resource file, so I've changed it.

tests/src/test/scala/system/basic/WskCliBasicTests.scala
steven0711dong: Nice catch!

commands/trigger.go
```diff
@@ -436,47 +440,118 @@ func (t *Trigger) Create(Client *whisk.Client, args []string) error {
...
 
+	//simplestTrigger indicates user are creating a trigger without any feed or parameters
+	simplestTrigger := len(Flags.trigger.feedParam) == 0 && len(Flags.common.param) == 0 && len(Flags.trigger.triggerParam) == 0
+
+	//if users are 1. creating a trigger without any feed or parameters
+	//             2. creating a trigger using --param flag
+	//then we use the old way to create the trigger.
+	if len(Flags.common.param) > 0 || simplestTrigger {
+		if len(Flags.trigger.feedParam) > 0 || len(Flags.trigger.triggerParam) > 0 {
+			whisk.Debug(whisk.DbgError, "User tries to mix use of --param with --feed-param and --trigger-param")
```
dubee: Is there a use case for the user trying to use trigger parameters and feed parameters together?

commands/trigger.go
```diff
@@ -436,47 +440,118 @@ func (t *Trigger) Create(Client *whisk.Client, args []string) error {
...
 	// if a feed is specified, add feed annotation the annotations declared on the command line
 	// TODO: add test to ensure that generated annotation has precedence
-	if feedName != nil {
-		annotationArray = append(annotationArray, getFormattedJSON("feed", feedName.GetFullQualifiedName()))
+	if feedQualifiedName != nil {
+		annotationArray = append(annotationArray, getFormattedJSON("feed", feedQualifiedName.GetFullQualifiedName()))
 	}
 	annotations := getParameters(annotationArray, true, true)
 
+	//simplestTrigger indicates user are creating a trigger without any feed or parameters
```
dubee: Does duplicate code exist in create and update? If so it'd be better to create functions out of the duplicate code to reduce complexity.

commands/commands.go
```diff
@@ -181,17 +184,34 @@ func parseArgs(args []string) ([]string, []string, []string, error) {
 					map[string]interface{}{"err": whiskErr})
 				whiskErr = whisk.MakeWskError(errors.New(errMsg), whisk.EXIT_CODE_ERR_GENERAL, whisk.DISPLAY_MSG,
 					whisk.DISPLAY_USAGE)
-				return nil, nil, nil, whiskErr
+				return nil, nil, nil, nil, nil, whiskErr
 			}
 		} else if args[i] == "-a" || args[i] == "--annotation" {
+			fmt.Println("4")
```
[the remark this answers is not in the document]
steven0711dong: oops..

commands/trigger.go
steven0711dong: Yeah, if users want to create a trigger using the new way. They will create trigger parameters using --trigger-param flag and create feed parameters using --feed-param flag. They just cannot combine --param with --feed-param or --trigger-param. The meaning is ambiguous. So we issue an error here.

commands/trigger.go
steven0711dong: I created two functions so that both create and update could use. Some other variables such as params, feed params do exist in both functions but are created differently and the creation depends on users' intention on how to issue trigger command(old way and new way). So it might make things more complicated to create functions out of duplicate code. 

I have to admit that after adding the new flags, both functions become bigger but I think it might be a good idea to separate execution path for the old way of creating trigger from the new way so they do not get lumped together.

commands/commands.go
dubee: Remove print statement.

commands/commands.go
```diff
@@ -19,6 +19,7 @@ package commands
 
 import (
 	"errors"
+	"fmt"
```
dubee: Remove `fmt`.

commands/trigger.go
```diff
@@ -436,47 +442,108 @@ func (t *Trigger) Create(Client *whisk.Client, args []string) error {
...
+		return nil
+	}
+
+	//if execution reached this line, it means that users intend to create or update trigger using the new way
+	//by using --feed-param or --trigger-param flag
+	if feedQualifiedName == nil && len(Flags.trigger.feedParam) > 0 {
+		//trigger without feed names cannot have a feed param
+		errMsg := fmt.Sprintf("Incorrect usage. trigger without a feed cannot have feed parameters. \n")
+		fmt.Fprintf(os.Stderr, "%s%s", color.RedString("error: "), errors.New(errMsg))
+		return nil
```
dubee: Hmm, errors are usually return back to the system instead of `nil`.

commands/trigger.go
```diff
@@ -597,7 +679,75 @@ func (t *Trigger) Update(Client *whisk.Client, args []string) error {
...
+//then we use the old way to create the trigger.
+func userIndicatesToUseOldTriggerCommand() bool {
+	return len(Flags.common.param) > 0 || (len(Flags.trigger.feedParam) == 0 && len(Flags.common.param) == 0 && len(Flags.trigger.triggerParam) == 0)
+}
+
+func userIssuedNewTriggerCmd() bool {
+	if len(Flags.trigger.feedParam) > 0 || len(Flags.trigger.triggerParam) > 0 {
+		whisk.Debug(whisk.DbgError, "User tries to mix use of --param with --feed-param and --trigger-param")
+		errMsg := fmt.Sprintf("Incorrect usage. Cannot combine --feed-param or --trigger-param flag with --param flag\n")
+		fmt.Fprintf(os.Stderr, "%s%s", color.RedString("error: "), errors.New(errMsg))
```
dubee: Errors are usually returned to the system, which halts further execution.

commands/trigger.go
```diff
@@ -436,47 +442,108 @@ func (t *Trigger) Create(Client *whisk.Client, args []string) error {
...
 	// if a feed is specified, add feed annotation the annotations declared on the command line
 	// TODO: add test to ensure that generated annotation has precedence
-	if feedName != nil {
-		annotationArray = append(annotationArray, getFormattedJSON("feed", feedName.GetFullQualifiedName()))
+	if feedQualifiedName != nil {
+		annotationArray = append(annotationArray, getFormattedJSON("feed", feedQualifiedName.GetFullQualifiedName()))
 	}
 	annotations := getParameters(annotationArray, true, true)
 
+	//Within this if statement, we process users' trigger command using the old way.
```
dubee: This is quite a large method with over 100 lines of code. Generally large functions are hard to follow and should be broken out into several functions.

commands/trigger.go
```diff
@@ -548,17 +623,30 @@ func (t *Trigger) Update(Client *whisk.Client, args []string) error {
...
 			map[string]interface{}{"name": qualifiedName.GetEntityName(), "err": err})
 		werr := whisk.MakeWskErrorFromWskError(errors.New(errStr), err, whisk.EXIT_CODE_ERR_GENERAL, whisk.DISPLAY_MSG, whisk.NO_DISPLAY_USAGE)
 		return werr
-	} else {
+	}
 
-		// Get full feed name from trigger get request as it is needed to get the feed
-		if retTrigger != nil && retTrigger.Annotations != nil {
-			fullFeedName = getValueString(retTrigger.Annotations, "feed")
+	// Get full feed name from trigger get request as it is needed to get the feed
```
dubee: This this update method has 200 lines of code making it extremely hard to understand quickly.

commands/trigger.go
steven0711dong: OK, I'll have both return error.

commands/trigger.go
steven0711dong: I can try to separate the 'old way' and 'new way' into two different functions.

steven0711dong:
I've already changed the error message thing, I'll look at how to break them into small functions tonight.

tests/src/test/scala/system/basic/WskCliBasicTests.scala
```diff
@@ -491,6 +491,47 @@ class WskCliBasicTests extends TestHelpers with WskTestHelpers {
...
+      trigger.create(triggerName, parameters = params)
+    }
+    wsk
+      .cli(
+        Seq("trigger", "update", triggerName, "-F", "feedParam", "feedParamVal", "--auth", wskprops.authKey) ++ wskprops.overrides,
+        expectedExitCode = ERROR_EXIT)
+      .stderr should include("this trigger does not contain a feed")
+  }
+
+  it should "return error message when creating or updating feed with both --param and --trigger-param/--feed-param flags" in withAssetCleaner(
```
dubee: Should there also be a test for update?

commands/trigger.go
```diff
@@ -547,15 +640,24 @@ func (t *Trigger) Update(Client *whisk.Client, args []string) error {
 	retTrigger, httpResp, err := Client.Triggers.Get(qualifiedName.GetEntityName())
 
 	if err != nil && httpResp.StatusCode == 404 {
-		t.Create(Client, args)
+		if createErr := t.Create(Client, args); createErr != nil {
```
dubee: This is new. If trigger get fails then trigger update now fails, correct?

steven0711dong: Correct. 
1. If trigger get fails with 404, then we create the trigger. If creating the triggers fails, we have to report back to the user. Previously we do not report any error when we create trigger in update and that was a bug. 
2. If trigger fails with other reason, execution will fall into next else if, and we report the error message as well.

commands/trigger.go
```diff
@@ -601,3 +703,110 @@ func (t *Trigger) Update(Client *whisk.Client, args []string) error {
...
+//             2. creating a trigger using --param flag
+//then we use the old way to create the trigger.
+func userIndicatesToUseOldTriggerCommand() bool {
+	return len(Flags.common.param) > 0 || (len(Flags.trigger.feedParam) == 0 && len(Flags.common.param) == 0 && len(Flags.trigger.triggerParam) == 0)
+}
+
+func userIssuedNewTriggerCmd() error {
+	if len(Flags.trigger.feedParam) > 0 || len(Flags.trigger.triggerParam) > 0 {
+		whisk.Debug(whisk.DbgError, "User tries to mix use of --param with --feed-param and --trigger-param")
+		errStr := wski18n.T("Incorrect usage. Cannot combine --feed-param or --trigger-param flag with --param flag")
```
dubee: Need to say `--param-file` to this list. Is this error message exported to wski18n?

commands/trigger.go
```diff
@@ -397,11 +397,15 @@ func init() {
 	triggerCreateCmd.Flags().StringSliceVarP(&Flags.common.param, "param", "p", []string{}, wski18n.T("parameter values in `KEY VALUE` format"))
 	triggerCreateCmd.Flags().StringVarP(&Flags.common.paramFile, "param-file", "P", "", wski18n.T("`FILE` containing parameter values in JSON format"))
 	triggerCreateCmd.Flags().StringVarP(&Flags.common.feed, "feed", "f", "", wski18n.T("trigger feed `ACTION_NAME`"))
+	triggerCreateCmd.Flags().StringSliceVarP(&Flags.trigger.feedParam, "feed-param", "F", []string{}, wski18n.T("feed parameter values in `KEY VALUE` format"))
```
dubee: Are the help entries added to wski18n?

tests/src/test/scala/system/basic/WskCliBasicTests.scala
dubee: @steven0711dong , I only see `create` tested here?

tests/src/test/scala/system/basic/WskCliBasicTests.scala
```diff
@@ -491,6 +491,47 @@ class WskCliBasicTests extends TestHelpers with WskTestHelpers {
...
+    wskprops) { (wp, assetHelper) =>
+    val triggerName = withTimestamp("t1tor1")
+    val ns = wsk.namespace.whois()
+
+    var stderr =
```
dubee: Test `-P` as well?

steven0711dong: I added a test for -P

commands/trigger.go
steven0711dong: added

commands/trigger.go
steven0711dong: made error msg more generic

wski18n/resources/en_US.all.json
```diff
@@ -1594,5 +1594,21 @@
...
+  {
+    "id": "trigger parameter values in `KEY VALUE` format", 
+    "translation": "trigger parameter values in `KEY VALUE` format"
+  },
+  {
+    "id": "feed parameter values in `KEY VALUE` format", 
+    "translation": "feed parameter values in `KEY VALUE` format"
+  },
+  {
+    "id": "Incorrect usage. trigger without a feed cannot have feed parameters",
```
dubee: Capital `t` in `trigger.

commands/trigger.go
```diff
@@ -431,52 +435,141 @@ func (t *Trigger) Create(Client *whisk.Client, args []string) error {
...
-		annotationArray = append(annotationArray, getFormattedJSON("feed", feedName.GetFullQualifiedName()))
+	if feedQualifiedName != nil {
+		annotationArray = append(annotationArray, getFormattedJSON("feed", feedQualifiedName.GetFullQualifiedName()))
 	}
+
 	annotations := getParameters(annotationArray, true, true)
 
+	//if trigger contains no feed but user tries to update feed parameter, then we issue error.
+	if feedQualifiedName == nil && len(Flags.trigger.feedParam) > 0 {
+		whisk.Debug(whisk.DbgError, "Incorrect usage. trigger without a feed cannot have feed parameters")
```
dubee: This error string does not match what is in `en_US.all.json`.

PR #480, opened 2020-01-28
require whisk auth fix

steven0711dong:


commands/action.go
```diff
@@ -849,49 +854,76 @@ func getWebSecureAnnotations(action *whisk.Action) whisk.KeyValueArr {
 	return webKvArr[0:j]
 }
 
+func getNewSecret(secret interface{}) string {
+	_, isJSONNum := secret.(json.Number)
```
rabbah, line 858: I'd replace this with the ability to generate an alphanumeric secret instead.
Something like:

```go
rand.Seed(time.Now().UnixNano())
var alphabet = []rune("abcdefghijklmnopqrstuvwxyzABCDEFGHIJKLMNOPQRSTUVWXYZ")
b := make([]rune, SECRET_LENGTH)
for i := range b {
   b[i] = alphabet[rand.Intn(len(alphabet))]
}
return string(b)
```

rabbah (commented):
I apologize no one reviewed this sooner. I added a comment about generating alphanumeric secrets instead of integer only. Also this needs either some go unit tests or integration (scala) tests.

PR #484, opened 2020-03-17
add -e param to pass environment values in KEY VALUE format

dilox:


dilox:
adding support for env variables at init [wip - do not merge]

rabbah:
@dilox i don't think this is sufficient. The way the cli handles `-p` and -`a` is with custom logic that has to be extended to `-e`. This will need some unit tests (or Scala integration tests).

Issue #481, opened 2020-02-06
wsk should remove trailing slash on apihost


travigd:
Just spent an embarrassing amount of time trying to determine why the wsk CLI was working on one machine and not on another. The ultimate issue was that I ended up specifying the `APIHOST` variable as `https://localhost:8443/` in the config file, with the trailing slash, which causes 404s since the CLI was sending requests to `https://localhost:8443//api/v1/namespaces/_/triggers?limit=0&skip=0` (note the double-slash after the domain).

rabbah:
Sorry about that @travigd. Agree this should get fixed. (Contributions welcomed :)

PR #485, opened 2020-03-17
remove trailing slash on apihost #481

dilox:


commands/property.go
```diff
@@ -471,6 +471,10 @@ func loadProperties() error {
 	}
 
 	if apiHost, hasProp := props["APIHOST"]; hasProp {
+		if last := len(apiHost) - 1; last >= 0 && apiHost[last] == '/' {
```
jiangpengcheng, line 474: how about using `Properties.APIHost = strings.TrimRight(apiHost, "/")` directly?

dilox: Ok, I'll change ASAP

rabbah:
Thanks @dilox for the contribution.

rabbah (approved):

PR #488, opened 2020-08-12
Support passing del annotation

ningyougang:
Currently, if passing another annotations, original previous annotation
will be removed and the passed new annotations will be added.

It may give users some confused that why my previous annotation gone.
So it is better to not delete user's previous annotation when adding new
annotation, but at the same time, need to provide a feature that
support to delete annotation by user via ClI, e.g.
```shell
wsk action update hello --del-annotation key1 --del-annotation key2
```

another brother prs to support del annotation: 
* https://github.com/apache/openwhisk-client-go/pull/137
* https://github.com/apache/openwhisk/pull/4940

### DIFF
--- a/commands/action.go
+++ b/commands/action.go
@@ -480,6 +480,9 @@ func parseAction(cmd *cobra.Command, args []string, update bool) (*whisk.Action,
 		return nil, noArtifactError()
 	}
 
+	if update {
+		action.DelAnnotations = Flags.action.delAnnotation
+	}
 	whisk.Debug(whisk.DbgInfo, "Parsed action struct: %#v\n", action)
 	return action, err
 }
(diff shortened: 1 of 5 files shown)


ningyougang:
build error
image
It seems need to merge this firstly: https://github.com/apache/openwhisk-client-go/pull/137 ?

commands/flags.go
```diff
@@ -136,22 +136,23 @@ type FlagsStruct struct {
...
+	memory        int
+	logsize       int
+	concurrency   int
+	result        bool
+	kind          string
+	main          string
+	url           bool
+	save          bool
+	saveAs        string
+	delAnnotation []string
```
ningyougang, line 155: Add delAnnotation field

rabbah:
This is great to finally have!

commands/action.go
```diff
@@ -1305,6 +1308,7 @@ func init() {
 	actionUpdateCmd.Flags().StringVarP(&Flags.common.paramFile, "param-file", "P", "", wski18n.T("`FILE` containing parameter values in JSON format"))
 	actionUpdateCmd.Flags().StringVar(&Flags.action.web, WEB_FLAG, "", wski18n.T("treat ACTION as a web action, a raw HTTP web action, or as a standard action; yes | true = web action, raw = raw HTTP web action, no | false = standard action"))
 	actionUpdateCmd.Flags().StringVar(&Flags.action.websecure, WEB_SECURE_FLAG, "", wski18n.T("secure the web action. where `SECRET` is true, false, or any string. Only valid when the ACTION is a web action"))
+	actionUpdateCmd.Flags().StringArrayVar(&Flags.action.delAnnotation, "del-annotation", []string{}, wski18n.T("del annotation"))
```
style95, line 1311:
```suggestion
	actionUpdateCmd.Flags().StringArrayVar(&Flags.action.delAnnotation, "del-annotation", []string{}, wski18n.T("the list of annotations to be deleted from the action"))
```

commands/action.go
```diff
@@ -480,6 +480,9 @@ func parseAction(cmd *cobra.Command, args []string, update bool) (*whisk.Action,
 		return nil, noArtifactError()
 	}
 
+	if update {
+		action.DelAnnotations = Flags.action.delAnnotation
```
style95, line 484: I am not clear about this, can users pass list of annotations? or they have to use `--del-annotation` for every annotation?

ningyougang: Must pass like this `--del-annotation key1 --del-annotation key2`, in go backend, it will recevie the value by array:
https://github.com/apache/openwhisk-client-go/pull/137/files#diff-7a2fb893e2fa76731b906c5689a14e95R39

commands/action.go
ningyougang: @style95 updated accordingly.

ningyougang:
The reason of the build failed is that
image

When get the openwhisk-client-go codes, it didn't fetch the latest codes.

ningyougang:
@rabbah ,have any way to fetch openwhisk-client-go's latest code when build this pr?

PR #489, opened 2020-08-18
Migrated to using go mod to manage dependencies

huydoan2:
This PR attempts to use `go mod`, a new technique, to manage dependencies.

We will address cleaning up gradle build code in a separate PR.


rabbah:
Should we even keep gradle build rather than use the go build?

rabbah:
Thanks for your contributions @huydoan2.

huydoan2:
> Should we even keep gradle build rather than use the go build?

I think it's simpler to keep one build system around. We could add a Makefile for convenience.

mrutkows:
@rabbah would love to remove Gradle... perhaps a follow-on PR...

PR #516, opened 2022-05-24
Support array result

ningyougang:
- [x] Support array result

Another brother prs:
https://github.com/apache/openwhisk-client-go/pull/153 (should be merged secondly)
https://github.com/apache/openwhisk-wskdeploy/pull/1153  (should be merged firslty)

This pr can be merged thirdly

After merged this three prs and if openwhisk itself support `support array result`:
- invoker supports array result: https://github.com/apache/openwhisk/pull/5290 (merged)
- runtime images should support array result as well
  - go: https://github.com/apache/openwhisk-runtime-go/pull/170
  - java:
  - ...

use wsk can get the correct arrary result. e.g
### invoke action
* Current is
image
* To-Be after the three prs merged
image
### activation get
* Currrent is
image
* To-Be after the three prs merged
image

### DIFF
--- a/commands/util.go
+++ b/commands/util.go
@@ -604,16 +604,20 @@ func getChildValueStrings(keyValueArr whisk.KeyValueArr, key string, childKey st
 	return res
 }
 
-func getValueFromJSONResponse(field string, response map[string]interface{}) interface{} {
+func getValueFromResponse(field string, response interface{}) interface{} {
 	var res interface{}
 
-	for key, value := range response {
-		if key == field {
-			res = value
-			break
+	if result, ok := response.(map[string]interface{}); ok {
+		for key, value := range result {
+			if key == field {
+				res = value
+				break
+			}
 		}
 	}
-
+	if result, ok := response.([]interface{}); ok {
+		res = result
+	}
 	return res
 }
 
--- a/commands/util.go
+++ b/commands/util.go
@@ -617,6 +617,8 @@ func getValueFromResponse(field string, response interface{}) interface{} {
 	}
 	if result, ok := response.([]interface{}); ok {
 		res = result
+	} else {
+		res = ""
 	}
 	return res
 }
--- a/commands/util.go
+++ b/commands/util.go
@@ -605,22 +605,18 @@ func getChildValueStrings(keyValueArr whisk.KeyValueArr, key string, childKey st
 }
 
 func getValueFromResponse(field string, response interface{}) interface{} {
-	var res interface{}
-
 	if result, ok := response.(map[string]interface{}); ok {
 		for key, value := range result {
 			if key == field {
-				res = value
-				break
+				return value
 			}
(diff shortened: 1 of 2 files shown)


commands/util.go
```diff
@@ -604,16 +604,20 @@ func getChildValueStrings(keyValueArr whisk.KeyValueArr, key string, childKey st
...
+	if result, ok := response.(map[string]interface{}); ok {
+		for key, value := range result {
+			if key == field {
+				res = value
+				break
+			}
 		}
 	}
-
+	if result, ok := response.([]interface{}); ok {
```
dgrove-oss, line 618: Should we insist that `field` be "" if response is an array?  I'm not sure exactly how this function is used, but it might be error prone to just silently return the array result when the caller asked for a specific field of what they must have been expecting to be a JSON object.

ningyougang: @dgrove-oss yes, you right.

But it is better to cherry this pr's commits to https://github.com/apache/openwhisk-cli/pull/517 and close this pr
I know it is not good due to `Fix deprecated issue(pull/517)` and this pr is target is different.

And due to dependency problem, it seems we must do this,
The reason is https://github.com/apache/openwhisk-cli/pull/517#issuecomment-1221219583

style95: @ningyougang 
This method was used to print the invocation result.
https://github.com/apache/openwhisk-cli/blob/92c251f5d3a873029ccd8d815409582301d1103d/commands/action.go#L238

Do we need to print the array result itself for the array case?

ningyougang: I tested in such case (btw, i stopped all invokers containers)
image

The result is nil
image

Finally is
image

If i added `sleep 61s` for that action and make all invokers up again, the result is
image

For normally hello-array action with this pr, the result is
image

ningyougang: Btw, i already cherry-pick this pr's commits to https://github.com/apache/openwhisk-cli/pull/517,
Because add-array-result feature is merged before than this pr, so when build this pr and execute this pr's test cases, it would execute openwhisk core relative's add-array-result feature test cases. e.g. when invoke below test case
image
it would invoke use `CLIActivationsOperations` to get result
image

When build openwhisk pr, it would use `RestActivationsOperations` to get the result

style95: ok, it seems now I get it.

Previously, it returns `var res interface{}` without any initialization, when it failed to find the `field`.
Is this an empty string in case it could not find anything?
Isn't it nil?

ningyougang: Yes, it would return emtry string now if it failed to find the `field`

style95: Was it an empty string or nil?

ningyougang: nil

ningyougang:
Closed as https://github.com/apache/openwhisk-cli/pull/517 merged

